Issue #87, opened 2019-08-09
Query params don't support string values that contain spaces


agandrabur:
When creating an OpenAPIv3 specs that has and endpoint with query params, it's not possible to pass a string that contains spaces

```yaml
...
paths:
  /api/test:
    get:
      operationId: testOperation
      summary: Test
      parameters:
        - in: query
          name: name
          schema:
            type: string
...
```
When calling the API with `curl -X GET "http://localhost:3030/api/test?name=test%20string -H "accept: application/vnd.api+json"`
the value in the handler function will be `r.ParamName="test"`

The issue seems to originate here: https://github.com/pace/bricks/blob/master/http/jsonapi/runtime/parameters.go#L93

PR #89, opened 2019-08-09
adds support for RFC3339 and time parameters #87

threez:
fixes the issue filed in #87 and additionally adds support for the type `time.Time`

### DIFF
--- a/http/jsonapi/runtime/parameters.go
+++ b/http/jsonapi/runtime/parameters.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"reflect"
 	"strings"
+	"time"
 )
 
 // ScanIn help to avoid missuse using iota for the possible values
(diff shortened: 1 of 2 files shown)


http/jsonapi/runtime/parameters.go
```diff
@@ -90,6 +91,33 @@ func ScanParameters(w http.ResponseWriter, r *http.Request, parameters ...*ScanP
...
+			if len(scanData) == 0 {
+				continue
+			}
+
+			// parse value
+			ti, err := time.Parse(time.RFC3339Nano, scanData)
+			if err != nil {
+				WriteError(w, http.StatusBadRequest, &Error{
+					Title:  fmt.Sprintf("invalid time value for %s", param.Name),
+					Detail: fmt.Sprintf("time value not in RFC3339 format got: %q", scanData),
```
Teelevision: RFC3339Nano is different from RFC3339 and thus should also be differentiated in the details.

http/jsonapi/runtime/parameters.go
```diff
@@ -90,6 +91,33 @@ func ScanParameters(w http.ResponseWriter, r *http.Request, parameters ...*ScanP
 			panic(fmt.Errorf("impossible scanning location: %d", param.Location))
 		}
 
+		// handle string and dates types
```
Teelevision: Documentation of the surrounding function `ScanParameters` needs to be updated, because it says

> The used scanning function is fmt.Sscan

which is not true anymore. But it is an implementation detail and thus should not be part of the documentation, I guess.

http/jsonapi/runtime/parameters_test.go
```diff
@@ -254,6 +255,62 @@ func TestScanNumericParametersInQueryFloatArrayFail(t *testing.T) {
 	}
 }
 
+func TestScanStringParameters(t *testing.T) {
+	req := httptest.NewRequest("GET", "/foo?value0=12%20asdd&value1=", nil)
```
Teelevision: What happens if using `+` instead of `%20`? Maybe we should also define whether that will be a space or not.

threez: Yes, it does, generally, the URL escaping rules apply. I added it to the test to make that clear.

http/jsonapi/runtime/parameters.go
threez: Updated the docs.

PR #106, opened 2019-09-09
Add support for additionalProperties

### LINKED ISSUE #61: Generator to support `additionalProperties`
### Problem

additionalProperties are not handled at the moment

### Proposed solution

```yaml
type: object
additionalProperties:
  type: string
```
Instead generate `map[string]string`.

```yaml
type: object
additionalProperties:
  type: integer
```
Instead generate `map[string]int`.

```yaml
type: object
additionalProperties: true
```

Instead generate `map[string]interface{}`.

In case properties are defined, a warning is displayed, that these are ignored, and only the map is generated.

### LINKED ISSUE #105: additionalProperties is not supported
OpenAPIv3 specs additionalProperties is not supported

```yaml
type: object
additionalProperties: true
```
This should generate a `map[string]interface{}`

```yaml
components:
  schemas:
    Messages:        # <---- dictionary
      type: object
      additionalProperties:
        $ref: '#/components/schemas/Message'
    Message:
      type: object
      properties:
        code:
          type: integer
        text:
          type: string
```
This should generate a `map[string]Message`

agandrabur:
Fixes #61 

Allows to define a generic `map[string] interface{}` by setting
```yaml
type: object
additionalProperties: true
```
or
```yaml
type: object
additionalProperties: {}
```

Allows to define map of objects `map[string] *Comment` by setting
```yaml
type:object
additionalProperties:
    $ref: "#/components/schemas/Comment"
```

Allows to define map of basic types `map[string] string` by setting
```yaml
type:object
additionalProperties:
    type: string
```
Other supported basic types:
* `integer` with formats: `int` (default, not specified), `int32`, `int64`
* `boolean`
* `number` with formats: `float32` and `float64` (default)

Warns if properties are defined


http/jsonapi/generator/generate_types.go
```diff
@@ -76,6 +76,48 @@ func (g *Generator) buildType(prefix string, stmt *jen.Statement, schema *openap
...
+						stmt.Int64()
+					}
+				case "number":
+					if val.AdditionalProperties.Value.Format == "float" {
+						stmt.Float32()
+					} else {
+						stmt.Float64()
+					}
+				case "string":
+					stmt.String()
```
threez: should the defauln from line 80 be de default here (`interface{}`)?

http/jsonapi/generator/internal/articles/open-api.json
```diff
@@ -226,6 +226,52 @@
...
+                }
+            },
+            "MapTypeString": {
+                "type": "object",
+                "additionalProperties": {
+                    "type": "string"
+                },
+                "properties": {
+                    "something": {
+                        "type": "string"
```
threez: what is with a date a string that formats a date? can we make use of the function defined here https://github.com/pace/bricks/blob/master/http/jsonapi/generator/generate_helper.go#L24 somehow?

http/jsonapi/generator/generate_types.go
agandrabur: when defined in yaml with `additionalProperties: true`, the value of `val.AdditionalProperties` is `nil`. The switch statement is unreachable in this case

threez: Sure, maybe more general, can we reuse the below mentioned func?

http/jsonapi/generator/internal/articles/open-api.json
agandrabur: this property was added just to test that a warning log is generated and the properties are ignored. It has nothing to do with a date format

http/jsonapi/generator/generate_types.go
agandrabur: Changed it to use goType()

http/jsonapi/generator/internal/articles/open-api.json
threez: it should be picked up by the goType func now. So it should be fine. I meant to maybe add another test case.

Issue #33, opened 2019-08-04
Health check that reports service status


threez:
Next to the health check that only returns OK we need to add a health check that checks it's dependencies.

* [ ] Check for redis / postgres availability
* [ ] New endpoint under `/health/service`
* [ ] cache the health result, don't allow dos via health (e.g. cache for 10sec)

*****

* [ ] TODO: Think about how a circuit breaker comes into play in the scenario.

PR #110, opened 2019-10-18
#33: added Healthchecker for services

SchlottiP:
Implemented healthchecker for services like redis and implemented the redis and postgres healthchecker


backend/redis/README.md
```diff
@@ -40,3 +40,7 @@
 * `REDIS_IDLE_CHECK_FREQUENCY` default: `1m`
     * Frequency of idle checks made by idle connections reaper. Default is 1 minute. -1 disables idle connections reaper, but idle connections are still discarded by the client if IdleTimeout is set.
     * Everything that can be parsed by [ParseDuration](https://golang.org/pkg/time/#ParseDuration)
+* `REDIS_HEALTH_KEY` default: `healthy`
+    * Name of the that is written to check, if redis is healthy
+* `POSTGRES_HEALTHCHECK_MAX_REQUEST_SEC` default: `10s`
```
threez: redis I assume

backend/redis/healthRedis.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (h *redisHealthCheck) Name() string {
+	return "redis"
+}
+
+// HealthCheck if the last result is outdated, redis is checked for writeability and readability,
+// otherwise return the old result
+func (h *redisHealthCheck) HealthCheck(currTime time.Time) error {
+	h.State.M.Lock()
```
threez: The state update and the lock around it should be part of the `h.State` and not leak to the users of the `h.State`

backend/redis/healthRedis.go
```diff
@@ -0,0 +1,56 @@
...
+	defer h.State.M.Unlock()
+	if currTime.Sub(h.State.LastCheck) > cfg.HealthMaxRequest {
+		//Set initial to healthy:
```
threez: check go style guide for comments

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,123 @@
...
+				// to increase performance of the request set
+				// content type and write status code explicitly
+				w.Header().Set("Content-Type", "text/plain")
+				w.WriteHeader(http.StatusOK)
+				w.Write([]byte("OK\n"[:])) // nolint: gosec,errcheck
+			} else {
+				w.Header().Set("Content-Type", "text/plain")
+				w.WriteHeader(http.StatusServiceUnavailable)
+				w.Write([]byte("Not OK:\n"[:])) // nolint: gosec,errcheck
+				w.Write([]byte(err.Error()[:])) // nolint: gosec,errcheck
```
Teelevision: I don't think we should expose the specific error. Imagine it would be something like "Password foobar is invalid."

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,123 @@
...
+			} else if healthy, err := hc.HealthCheck(time.Now()); healthy {
+				// to increase performance of the request set
+				// content type and write status code explicitly
+				w.Header().Set("Content-Type", "text/plain")
+				w.WriteHeader(http.StatusOK)
+				w.Write([]byte("OK\n"[:])) // nolint: gosec,errcheck
+			} else {
+				w.Header().Set("Content-Type", "text/plain")
+				w.WriteHeader(http.StatusServiceUnavailable)
+				w.Write([]byte("Not OK:\n"[:])) // nolint: gosec,errcheck
```
Teelevision: I imagine someone naively implementing a health check like `curl .../health/foobar | grep OK`. So maybe return something like "FAIL" or "ERROR".

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,123 @@
...
+				w.Write([]byte(err.Error()[:])) // nolint: gosec,errcheck
+			}
+		}
+	} else {
+		w.Header().Set("Content-Type", "text/plain")
+		w.WriteHeader(http.StatusBadRequest)
```
Teelevision: I think `404 Not Found` would be better suited for a non-existing health check.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,123 @@
...
+	route := r.URL.Path
+	splitRoute := strings.Split(route, "/health/")
+	if len(splitRoute) == 2 && splitRoute[0] == "" && checks[splitRoute[1]] != nil {
+		hc := checks[splitRoute[1]]
+		if splitRoute[1] == hc.Name() {
+			//If it was not possible to initialise this health check, then show the initialisation error
+			if err := initErrors[hc.Name()]; err != nil {
+				w.Header().Set("Content-Type", "text/plain")
+				w.WriteHeader(http.StatusServiceUnavailable)
+				w.Write([]byte("Not OK:\n"[:])) // nolint: gosec,errcheck
```
Teelevision: Why `[:]`?

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,123 @@
...
+func (cs *ConnectionState) setConnectionState(healthy bool, err error, mom time.Time) {
+	cs.m.Lock()
+	defer cs.m.Unlock()
+	cs.isHealthy = healthy
+	cs.err = err
+	cs.LastCheck = mom
+}
+
+func (cs *ConnectionState) SetErrorState(err error, mom time.Time) {
+	cs.setConnectionState(false, err, mom)
```
Teelevision:
```suggestion
	cs.setConnectionState(err == nil, err, mom)
```

For example in the redis health check you do this:

```go
errRead := h.client.Get(cfg.HealthKey).Err()
h.State.SetErrorState(errRead, currTime)
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,123 @@
...
+	if router == nil {
+		log.Warnf("Tried to add HealthCheck ( %T ) without a router", hc)
+		return
+	}
+	checks[hc.Name()] = hc
+	if err := hc.InitHealthCheck(); err != nil {
+		log.Warnf("Error initialising HealthCheck  %T: %v", hc, err)
+		initErrors[hc.Name()] = err
+	}
+	router.Handle("/health/"+hc.Name(), Handler())
```
Teelevision: Note somewhere that all calls to `RegisterHealthCheck` should be done before serving the router. Otherwise you can get a race condition with writing to `checks` while at the same time reading it because a request was made to the health endpoint. Or even better use a `sync.RWLock` to prevent this in general.

maintenance/health/servicehealthcheck/README.md
```diff
@@ -0,0 +1,5 @@
+# Health Checks for Services 
+* makes it possible to have a health check for different services (e.g. postgres, redis)
+* A service has to implement the HealthCheck interface and register the healthCheck with `healthchecker.RegisterHealthCheck(hc HealthCheck)`
+* `healthchecker` does NOT make any caching of the results but it offers a struct (`ConnectionState`) for caching
+ the result in the Implementations of the Health Checks
```
Teelevision: Please straighten this readme. Like indentions, usage of backticks (`), capitalised letters. Also "does NOT cache the results".

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,123 @@
...
+func NewConnectionState() *ConnectionState {
+	return &ConnectionState{m: sync.Mutex{}}
+}
+
+func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	route := r.URL.Path
+	splitRoute := strings.Split(route, "/health/")
+	if len(splitRoute) == 2 && splitRoute[0] == "" && checks[splitRoute[1]] != nil {
+		hc := checks[splitRoute[1]]
+		if splitRoute[1] == hc.Name() {
```
Teelevision: I would find it rather strange that the name can change afterwards. But any object implementing the interface is allowed to do so. See my other suggesting that would get rid of the `Name()` method.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,123 @@
...
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+)
+
+type HealthCheck interface {
+	Name() string
+	InitHealthCheck() error
+	HealthCheck(currTime time.Time) (bool, error)
```
Teelevision: Why is it passing the current time? It's not like the current time can be anything else than `time.Now()`, logically. It also implies that a health check is doing throttling, which should be completely the responsibility of the health check. Or worse, it could imply a health check should be able to pull up results from the past, i.e. provide a full history.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,138 @@
...
+		log.Debugf("Health checks can only be added once, tried to add another health check with name %v", name)
+		return
+	}
+	if router == nil {
+		log.Warnf("Tried to add HealthCheck ( %T ) without a router", hc)
+		return
+	}
+	checksLock.Lock()
+	defer checksLock.Unlock()
+	checks[name] = hc
```
Teelevision: The `checksLock.Unlock()` is deferred, so it is called after `hc.InitHealthCheck()` has run. But initialising a health check can take some time. Because of the lock all health checks are initialised one after another. We would be better off if those could run concurrently. So I'd suggest you don't defer the unlock and instead call it directly after `checks[name] = hc`.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,138 @@
...
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(errorCode)
+	w.Write([]byte("Error \n")) // nolint: gosec,errcheck
+	log.Warnf("helthcheck %v was not healthy %v", name, err)
+}
+
+// RegisterHealthCheck register a healthCheck that need to be routed
+// names must be uniq
+func RegisterHealthCheck(hc HealthCheck, name string) {
+	if checks[name] != nil {
```
Teelevision: Surround with read lock. Also we usually do such a check like this: `if _, ok := checks[name]; ok { // exists`

Teelevision: Just an idea, you could add a type for the `checks`, like this:

```go
type healthChecks struct {
	checks map[string]HealthCheck
	m      sync.RWMutex
}

func (h *healthChecks) Set(name string, hc HealthCheck) {
	h.m.Lock()
	defer h.m.Unlock()
	h.checks[name] = hc
}

func (h *healthChecks) Get(name string) HealthCheck {
	h.m.RLock()
	defer h.m.RUnlock()
	return h.checks[name]
}

var checks = healthChecks{
	checks: make(map[string]HealthCheck),
}
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+type handler struct{}
+
+// checks map with all health checks, key: Name of the check
+var checks = make(map[string]HealthCheck)
+var checksLock = sync.RWMutex{}
+
+// initErrors map with all errors that happened in the initialisation of the health checks
+var initErrors = make(map[string]error)
```
Teelevision: `initErrors` needs a lock just as `checksLock`

backend/redis/health_redis.go
```diff
@@ -0,0 +1,48 @@
...
+	h.State = servicehealthcheck.NewConnectionState()
+	return nil
+}
+
+// HealthCheck checks if the redis is healthy. If the last result is outdated, redis is checked for writeability and readability,
+// otherwise return the old result
+func (h *redisHealthCheck) HealthCheck() (bool, error) {
+	currTime := time.Now()
+	if currTime.Sub(h.State.LastCheck) > cfg.HealthMaxRequest {
+		h.State.SetHealthy(currTime)
```
Teelevision: If there is a second call to `HealthCheck()` while the first one is being processed, the state is reported back as healthy, but in fact it should be the previous status. This is because this line sets the state to healthy without the check being processed. The second call will skip the test because of `cfg.HealthMaxRequest` and report it being healthy.

backend/postgres/postgres.go
```diff
@@ -117,9 +122,11 @@ func init() {
 	if err != nil {
 		log.Fatalf("Failed to parse postgres environment: %v", err)
 	}
+	servicehealthcheck.RegisterHealthCheck(&PgHealthCheck{}, "postgresdefault")
+
 }
 
-// ConnectionPool returns a new database connection pool
+// ConnectionPool returns a new database connection Pool
```
Teelevision: Huh?

backend/postgres/health_postgres.go
```diff
@@ -0,0 +1,52 @@
...
+	currTime := time.Now()
+	if currTime.Sub(h.State.LastCheck) > cfg.HealthMaxRequest {
+		h.State.SetHealthy(currTime)
+		//Readcheck
+		_, errRead := h.Pool.Exec(`SELECT 1; `)
+		if errRead != nil {
+			h.State.SetErrorState(errRead, currTime)
+		} else {
+			//writecheck - create Table if not exist and add Data
+			_, errWrite := h.Pool.Exec(`CREATE TABLE IF NOT EXISTS ` + cfg.HealthTableName + `(ok boolean);`)
```
threez: can we move this to the `InitHealthCheck`

backend/redis/redis.go
```diff
@@ -7,15 +7,16 @@ package redis
 import (
 	"context"
 	"fmt"
+	"github.com/pace/bricks/maintenance/health/servicehealthcheck"
```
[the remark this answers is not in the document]
threez: please additionally setup go imports like formatting, to prevent this in the future

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,138 @@
...
+	return &ConnectionState{m: sync.Mutex{}}
+}
+
+func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	route := r.URL.Path
+	splitRoute := strings.Split(route, "/health/")
+
+	//Check the route first
+	if len(splitRoute) < 2 || splitRoute[0] != "" {
+		h.writeError(w, errors.New("Route not Valid\n"), http.StatusBadRequest, route)
```
threez:
```suggestion
		h.writeError(w, errors.New("Route not valid"), http.StatusBadRequest, route)
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,138 @@
...
+		return
+	}
+
+	checksLock.RLock()
+	defer checksLock.RUnlock()
+	name := splitRoute[1]
+
+	if checks[name] == nil {
+		h.writeError(w, errors.New("Health Check not registered\n"), http.StatusNotFound, name)
```
threez:
```suggestion
		h.writeError(w, errors.New("Health check not registered"), http.StatusNotFound, name)
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,138 @@
...
+		w.Write([]byte("OK\n")) // nolint: gosec,errcheck
+	} else {
+		h.writeError(w, err, http.StatusServiceUnavailable, name)
+	}
+}
+
+func (h *handler) writeError(w http.ResponseWriter, err error, errorCode int, name string) {
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(errorCode)
+	w.Write([]byte("Error \n")) // nolint: gosec,errcheck
```
threez:
```suggestion
	w.Write([]byte("ERR")) // nolint: gosec,errcheck
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// RegisterHealthCheck register a healthCheck that need to be routed
+// names must be uniq
+func RegisterHealthCheck(hc HealthCheck, name string) {
+	if checks[name] != nil {
+		log.Debugf("Health checks can only be added once, tried to add another health check with name %v", name)
+		return
+	}
+	if router == nil {
```
threez: don't do nil checking, why would it be nil, isn't this a panic?

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,138 @@
...
+		log.Warnf("Tried to add HealthCheck ( %T ) without a router", hc)
+		return
+	}
+	checksLock.Lock()
+	defer checksLock.Unlock()
+	checks[name] = hc
+	if err := hc.InitHealthCheck(); err != nil {
+		log.Warnf("Error initialising HealthCheck  %T: %v", hc, err)
```
threez:
```suggestion
		log.Warnf("Error initialising health check  %T: %v", hc, err)
```

maintenance/health/servicehealthcheck/healthchecker_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+}
+
+func TestHandlerOK(t *testing.T) {
+	setup(&testHealthChecker{name: "test"})
+	if resp.StatusCode != 200 {
+		t.Errorf("Expected /health to respond with 200, got: %d", resp.StatusCode)
+	}
+
+	data, err := ioutil.ReadAll(resp.Body)
```
threez: create a test helper to DRY up the code

maintenance/health/servicehealthcheck/healthchecker.go
SchlottiP: used sync.Map instead

maintenance/health/servicehealthcheck/healthchecker.go
SchlottiP: initErrors is now sync.Map

maintenance/health/servicehealthcheck/healthchecker.go
SchlottiP: O don't think so because then health checks could make an application unusable. Changed the logging Level to Error.

maintenance/health/servicehealthcheck/healthchecker_test.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil
+	}
+}
+
+func (t *testHealthChecker) HealthCheck() (bool, error) {
+	if t.healthCheckErr {
+		return false, errors.New("healtherror")
+	} else {
+		return true, nil
+	}
```
Teelevision: go-lint says:
> if block ends with a return statement, so drop this else and outdent its block

Same above.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,125 @@
...
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/pace/bricks/maintenance/errors"
+	"github.com/pace/bricks/maintenance/log"
+)
+
+type HealthCheck interface {
```
Teelevision: Add a comment to document this type. `// HealthCheck` ...

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (cs *ConnectionState) SetHealthy(mom time.Time) {
+	cs.setConnectionState(true, nil, mom)
+}
+
+func (cs *ConnectionState) GetState() (bool, error) {
+	return cs.isHealthy, cs.err
+}
+
+func NewConnectionState() *ConnectionState {
```
Teelevision: Document this function.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (cs *ConnectionState) setConnectionState(healthy bool, err error, mom time.Time) {
+	cs.m.Lock()
+	defer cs.m.Unlock()
+	cs.isHealthy = healthy
+	cs.err = err
+	cs.LastCheck = mom
+}
+
+func (cs *ConnectionState) SetErrorState(err error, mom time.Time) {
```
Teelevision: Add comments to document all the exported methods.

maintenance/health/servicehealthcheck/healthchecker.go
Teelevision: Also go-lint says:
> error strings should not be capitalized or end with punctuation or a newline

maintenance/health/servicehealthcheck/healthchecker_test.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (t *testHealthChecker) CleanUp() error {
+	return nil
+}
+
+var resp *http.Response
```
Teelevision: Tests can run concurrently as well. So you should not use global state.

backend/postgres/health_postgres.go
```diff
@@ -0,0 +1,56 @@
...
+package postgres
+
+import (
+	"time"
+
+	"github.com/go-pg/pg"
+	"github.com/pace/bricks/maintenance/health/servicehealthcheck"
+)
+
+type PgHealthCheck struct {
```
Teelevision: Add comments to this type and its exported methods for documentation.

Teelevision:
@Smuerdt can you review, please? We should probably squash this one when merging.

backend/postgres/postgres.go
```diff
@@ -62,6 +63,12 @@ type config struct {
 	// but idle connections are still discarded by the client
 	// if IdleTimeout is set.
 	IdleCheckFrequency time.Duration `env:"POSTGRES_IDLE_CHECK_FREQUENCY" envDefault:"1m"`
+	// Name of the Table that is created to try if database is writeable
+	HealthTableName string `env:"POSTGRES_HEALTHCHECK_TABLE" envDefault:"healthcheck"`
```
Smuerdt: Maybe streamline naming here and for `HealthMaxRequest` and `HealthWrite`. Example: 
```go
HealthCheckTableName string `env:"POSTGRES_HEALTH_CHECK_TABLE_NAME" envDefault:"healthcheck"`
```

backend/redis/health_redis.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/go-redis/redis"
+	"github.com/pace/bricks/maintenance/health/servicehealthcheck"
+)
+
+// HealthCheck checks the state of a redis connection. It must not be changed
+// after it was registered as a health check.
+type HealthCheck struct {
+	state      servicehealthcheck.ConnectionState
+	Client     *redis.Client
+	CheckWrite bool
```
Smuerdt: Why is this calles `CheckWrite` when the default healthcheck in `HealthCheck()` is always writing?

backend/redis/redis.go
```diff
@@ -34,6 +35,12 @@ type config struct {
 	PoolTimeout        time.Duration `env:"REDIS_POOL_TIMEOUT"`
 	IdleTimeout        time.Duration `env:"REDIS_IDLE_TIMEOUT"`
 	IdleCheckFrequency time.Duration `env:"REDIS_IDLE_CHECK_FREQUENCY"`
+	// Name of the key that is written to check, if redis is healthy
+	HealthKey string `env:"REDIS_HEALTH_KEY" envDefault:"healthy"`
+	// Amount of time to cache the last health check result
+	HealthMaxRequest time.Duration `env:"REDIS_HEALTHCHECK_MAX_REQUEST_SEC" envDefault:"10s"`
```
Smuerdt: Same as in `backend/postgres/postgres.go`: Think about streamlining naming here.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+type Initialisable interface {
+	Init() error
+}
+
+type handler struct{}
+
+// checks contains all registered Health Checks - key:Name
+var checks sync.Map
```
Smuerdt: Why not move these variables into the `handler` struct?

backend/postgres/health_postgres.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/go-pg/pg"
+	"github.com/pace/bricks/maintenance/health/servicehealthcheck"
+)
+
+// HealthCheck checks the state of a postgres connection. It must not be changed
+// after it was registered as a health check.
+type HealthCheck struct {
+	state      servicehealthcheck.ConnectionState
+	Pool       *pg.DB
+	CheckWrite bool
```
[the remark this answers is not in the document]
Teelevision: That's why the documentation of this type states that you must not change the health check after it is registered. Even if someone would do it, it would be a false negative, which I think is ok for a configuration issue.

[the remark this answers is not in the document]
Smuerdt: Ok, I'm fine with that but still would try to avoid things like this in the future.

maintenance/health/servicehealthcheck/healthchecker.go
Ferlonas: I don't see any improvement in doing this, because then we would have to store the handler as package variable (we need access to initErrors during health check registration). Resolving this as discussed with @Smuerdt

Issue #86, opened 2019-08-05
Add counter for panics


threez:
## Problem

After a new deployment of the software, we may see an increase of panics.

## Solution

All error handlers should increase the panic counter, that can then be centrally checked

threez:
@gabrielruiu nice, thanks for your PR, see the review comments.

PR #112, opened 2019-11-04
#86 - add prometheus gauge to count how many panics are caught within…

gabrielruiu:
… http request handlers

### DIFF
--- a/http/metrics.go
+++ b/http/metrics.go
@@ -52,16 +52,11 @@ var (
 		},
 		[]string{"code", "method", "source"},
 	)
-
-	paceHTTPPanicCounter = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "pace_http_panic_total",
-		Help: "A counter for panics intercepted while handling a request",
-	})
 )
 
 func init() {
 	// Register all of the metrics in the standard registry.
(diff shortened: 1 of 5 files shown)


http/metrics.go
```diff
@@ -52,11 +52,16 @@ var (
 		},
 		[]string{"code", "method", "source"},
 	)
+
+	paceHTTPPanicCounter = prometheus.NewGauge(prometheus.GaugeOpts{
```
threez: could you add the new metric to the readme?

http/router.go
```diff
@@ -23,7 +23,7 @@ func Router() *mux.Router {
 	r.Use(metricsMiddleware)
 
 	// last resort error handler
-	r.Use(errors.Handler())
+	r.Use(errors.HandlerWithCounter(paceHTTPPanicCounter))
```
threez: Can we maintain the `errors.Handler` interface for backwards compatibility. I don't think we need both forms.

gabrielruiu: Added a modification such that the `errors.Handler()`  stays the same

http/metrics.go
gabrielruiu: added reference to the metric in the `metrics/README.md`, but it feels kinds of out-of-place. If there is a better place to put it, let me know

Issue #109, opened 2019-10-17
Get rid of oauth2 middleware and generate the code instead


threez:
### Problem

Changing the authentication schema in the openapiv3 template currently has no effect on the authentication. The authentication is therefore hard to maintain from a single point and documentation and implmentation deviate.

### Solution

Generate the authentication code instead of using the oauth2 middleware directly.

PR #116, opened 2019-11-08
#109: Add the Authentication to the generated Api code

SchlottiP:
Adds the Authentication for ApiKey and OAuth2 to the generated ApiKey. 
This makes it necessary that each Api-Definition contains security schema configurations. But the generated code is still backwards compatible, because it can be used without the usage of the generated Authentication.


http/jsonapi/generator/internal/fueling/open-api_test.go
```diff
@@ -536,17 +539,25 @@ type Service interface {
...
+	s1.Methods("GET").Path("/beta/gas-stations/{gasStationId}/pumps/{pumpId}").Handler(GetPumpHandler(service, authBackend)).Name("GetPump")
+	s1.Methods("POST").Path("/beta/gas-station/{gasStationId}/payment").Handler(ProcessPaymentHandler(service, authBackend)).Name("ProcessPayment")
+	s1.Methods("POST").Path("/beta/gas-stations/{gasStationId}/approaching").Handler(ApproachingAtTheForecourtHandler(service, authBackend)).Name("ApproachingAtTheForecourt")
 	return router
 }
+
+// Router kept for backward compatibility. Please use RouteWithAuthentication
+func Router(service Service) *mux.Router {
+	return RouterWithAuthentication(service, nil)
+}
```
Teelevision: Just leaving a note here that, unlike what I previous said, we should keep this `Router` method.

http/jsonapi/generator/internal/pay/open-api_test.go
```diff
@@ -110,9 +132,20 @@ func GetPaymentMethodsHandler(service Service) http.Handler {
 CreatePaymentMethodSEPAHandler handles request/response marshaling and validation for
  Post /beta/payment-methods/sepa-direct-debit
 */
-func CreatePaymentMethodSEPAHandler(service Service) http.Handler {
+func CreatePaymentMethodSEPAHandler(service Service, authBackend AuthenticationBackend) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		defer errors.HandleRequest("CreatePaymentMethodSEPAHandler", w, r)
+		// Only do Authentication if a authentication Backend is available. Otherwise this is handled somewhere else
+		if authBackend != nil {
```
Teelevision: We should not skip authentication. Having no authentication backend is a configuration problem and should not cause an authentication bypass issue.

Teelevision:
I got some code that previous was generated like this:
```go
// Scan and validate incoming request parameters
vars := mux.Vars(r)
if !runtime.ScanParameters(w, r, &runtime.ScanParameter{
	Data:     &request.ParamAcceptLanguage,
	Location: runtime.ScanInHeader,
	Input:    vars["Accept-Language"],
	Name:     "Accept-Language",
}) {
	return
}
```
And now it's:
```go
// Scan and validate incoming request parameters
if !runtime.ScanParameters(w, r, &runtime.ScanParameter{
	Data:     &request.ParamAcceptLanguage,
	Location: runtime.ScanInHeader,
	Name:     "Accept-Language",
}) {
	return
}
```
Definition looks like:
```json
"parameters": [
  {
    "in": "header",
    "name": "Accept-Language",
    "schema": {
      "type": "string"
    }
  }
],
```

SchlottiP:
> I got some code that previous was generated like this:
> 
> ```go
> // Scan and validate incoming request parameters
> vars := mux.Vars(r)
> if !runtime.ScanParameters(w, r, &runtime.ScanParameter{
> 	Data:     &request.ParamAcceptLanguage,
> 	Location: runtime.ScanInHeader,
> 	Input:    vars["Accept-Language"],
> 	Name:     "Accept-Language",
> }) {
> 	return
> }
> ```
> 
> And now it's:
> 
> ```go
> // Scan and validate incoming request parameters
> if !runtime.ScanParameters(w, r, &runtime.ScanParameter{
> 	Data:     &request.ParamAcceptLanguage,
> 	Location: runtime.ScanInHeader,
> 	Name:     "Accept-Language",
> }) {
> 	return
> }
> ```
> 
> Definition looks like:
> 
> ```json
> "parameters": [
>   {
>     "in": "header",
>     "name": "Accept-Language",
>     "schema": {
>       "type": "string"
>     }
>   }
> ],
> ```

The Scan Parameter never generated anything for the Input fields for parameters from header (see `generate_handler.go` and the tests for fueling-api generation in the same package) : 
```
if param.Value.In == "path" {
									g.Id("Input").Op(":").Id("vars").Index(jen.Lit(param.Value.Name)).Op(",")
								}
```

http/jsonapi/generator/internal/pay/open-api_test.go
Teelevision: This is still open.

http/oauth2/authenticator.go
```diff
@@ -0,0 +1,64 @@
...
+// Created at 2019/10/30 by Charlotte Pröller
+
+package oauth2
+
+import (
+	"context"
+	"net/http"
+)
+
+// Authenticator is a Implementation of security.Authenticator for oauth2
```
Teelevision:
```suggestion
// Authenticator is an implementation of security.Authenticator for oauth2
```

http/oauth2/oauth2.go
```diff
@@ -101,94 +116,78 @@ func fromIntrospectResponse(s *IntrospectResponse, tokenValue string) token {
...
-		return nil
-	}
-
-	return val.(*token)
-}
-
-// WithBearerToken returns a new context that has the given bearer token set.
-// Use BearerToken() to retrieve the token. Use Request() to obtain a request
-// with the Authorization header set accordingly.
-func WithBearerToken(ctx context.Context, bearerToken string) context.Context {
```
Teelevision: Previously exported function `WithBearerToken` should be kept.

http/oauth2/oauth2.go
```diff
@@ -101,94 +116,78 @@ func fromIntrospectResponse(s *IntrospectResponse, tokenValue string) token {
...
 	}
 
-	authHeaderVal := headerPrefix + bt
+	authHeaderVal := security.HeaderPrefix + bt
 	r.Header.Set("Authorization", authHeaderVal)
 	return r
 }
 
-// BearerToken returns the bearer token stored in ctx
-func BearerToken(ctx context.Context) (string, bool) {
```
Teelevision: Previously exported function `BearerToken` should be kept.

http/jsonapi/generator/internal/pay/open-api.json
```diff
@@ -1159,6 +1171,30 @@
...
+              "id:sessions:create": "Create a new browser session"
+            }
+          }
+        }
+      },
+      "ProfileKey": {
+        "type": "apiKey",
+        "in": "header",
+        "name": "Authorization",
+        "description": "prefix with Bearer "
```
Teelevision:
```suggestion
        "description": "prefix with \"Bearer \""
```

http/security/helper.go
```diff
@@ -0,0 +1,70 @@
...
+
+type bearerToken struct {
+	value string
+}
+
+type Ctxkey string
+
+const HeaderPrefix = "Bearer "
+
+var TokenKey = Ctxkey("Token")
```
Teelevision: `Ctxkey` and `TokenKey` should not be exported. See https://golang.org/pkg/context/#Context

> A key identifies a specific value in a Context. Functions that to store values in Context typically allocate a key in a variable then use that key as the argument to context.WithValue and Context.Value. A key can be any type that supports equality; packages should define keys as an unexported type to avoid collisions.

maintenance/errors/error.go
```diff
@@ -10,7 +10,7 @@ import (
 	"net/http"
 	"os"
 
-	raven "github.com/getsentry/raven-go"
+	"github.com/getsentry/raven-go"
```
Teelevision: Even though the `raven` name is not necessary, if it differs from the name in the import path, you should still specify it. This is for readability.

http/jsonapi/generator/internal/pay/open-api_test.go
SchlottiP: the check is needed, because we still offer the old Router Method for initialisation, that does not pass an authBackend

http/security/helper.go
```diff
@@ -0,0 +1,76 @@
...
+		http.Error(w, "Unauthorized", http.StatusUnauthorized)
+		return ""
+	}
+	tokenValue := items[1]
+	return tokenValue
+}
+
+// ContextWithTokenKey creates a context with a Token
+func ContextWithTokenKey(targetCtx context.Context, token Token) context.Context {
+	return context.WithValue(targetCtx, tokenKey, token)
```
Teelevision: `targetCtx`: why target? I think just `ctx` would be fine.

http/security/helper.go
```diff
@@ -0,0 +1,76 @@
...
+
+package security
+
+import (
+	"context"
+	"net/http"
+	"strings"
+)
+
+type Token interface {
```
Teelevision: Add type documentation.

http/security/helper.go
```diff
@@ -0,0 +1,76 @@
...
+	token := TokenFromContext(ctx)
+
+	if token == nil {
+		return "", false
+	}
+
+	return token.GetValue(), true
+}
+
+func TokenFromContext(ctx context.Context) Token {
```
Teelevision: Add documentation.

Tip: You can find this type of error easily using [golint](https://github.com/golang/lint).

http/security/helper.go
```diff
@@ -0,0 +1,76 @@
...
+
+// GetValue returns the be
+func (b *bearerToken) GetValue() string {
+	return b.value
+}
+
+// GetBearerTokenFromHeader get the bearer Token from the header of the request
+// success: returns the bearer token without the Prefix
+// error: write the error in the response and return a empty bearer
+func GetBearerTokenFromHeader(r *http.Request, w http.ResponseWriter, headerName string) string {
```
Teelevision: I'm confused about all these methods: `GetBearerTokenFromHeader`, `ContextWithTokenKey`, `TokenFromContext`, `WithBearerToken` and `BearerToken`.

There is no harmony in this. `WithBearerToken` and `BearerToken` are very symmetrical, but their naming is totally different from the very similar functions `ContextWithTokenKey` and `TokenFromContext`. `GetBearerTokenFromHeader` has no equivalent setter.

When should I use `ContextWithTokenKey` and when `WithBearerToken`? It seems redundant to me having both functions.

`GetBearerTokenFromHeader` is not a getter, it might actually call `http.Error(w, "Unauthorized", http.StatusUnauthorized)`. But the name suggests that it is only a getter.

I would suggest that you reduce the number of functions, for example:
```go
func WithToken(context.Context, Token) context.Context {}
func GetToken(context.Context) Token {}
```

I'm not sure what to do about the `http.Error(w, "Unauthorized", http.StatusUnauthorized)`. It seems so be misplaced in `GetBearerTokenFromHeader` but I'm not sure where it belongs. Making it part of the `Authenticator`'s `Authorize` method could work.

http/security/authenticator.go
```diff
@@ -0,0 +1,18 @@
...
+package security
+
+import (
+	"context"
+	"net/http"
+)
+
+// Authenticator describes the needed functions for a authentication,
+// already implemented in oauth2.Authenticator and apikey.Authenticator
+type Authenticator interface {
```
Teelevision: If it's only method is `Authorize`, shouldn't the type actually be called `Authorizer` instead of `Authenticator`?

http/security/apikey/authorizer.go
```diff
@@ -0,0 +1,61 @@
...
+		log.Req(r).Info().Msg(err.Error())
+		http.Error(w, "Unauthorized", http.StatusUnauthorized)
+		return r.Context(), false
+	}
+	if key == a.apiKey {
+		return security.ContextWithTokenKey(r.Context(), &bearerToken{key}), true
+	}
+	err = errors.New("ApiKey not valid")
+	log.Req(r).Info().Msg(err.Error())
+	http.Error(w, err.Error(), http.StatusUnauthorized)
```
Teelevision: You can use a simple string here for `err`.

http/security/apikey/authorizer_test.go
```diff
@@ -0,0 +1,87 @@
...
+	_, b := auth.Authorize(r, w)
+
+	resp := w.Result()
+	body, err := ioutil.ReadAll(resp.Body)
+	defer resp.Body.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !b {
+		t.Errorf("Expected no error in Authentication, but was not succesfull with code %d and body %v", resp.StatusCode, string(body))
```
Teelevision:
```suggestion
		t.Errorf("Expected no error in authentication, but failed with code %d and body %v", resp.StatusCode, string(body))
```

http/jsonapi/generator/generate_handler.go
```diff
@@ -565,6 +588,38 @@ func (g *Generator) buildHandler(method string, op *openapi3.Operation, pattern
 	return route, nil
 }
 
+func generateAuthentication(op *openapi3.Operation, secSchemes map[string]*openapi3.SecuritySchemeRef) (*jen.Group, error) {
+	r := &jen.Group{}
+	r.Add(jen.Comment("Authentication Handling "))
+	for _, sl := range *op.Security {
+		for name, val := range sl {
+			securityScheme := secSchemes[name]
+			switch securityScheme.Value.Type {
```
Teelevision: I think it would be best to add a default case to the switch in case we encounter an unsupported type. Like maybe panic then.

http/oauth2/oauth2_test.go
```diff
@@ -80,6 +81,143 @@ func TestHandlerIntrospectError(t *testing.T) {
...
+	}
+	for _, tC := range testCases {
+		t.Run(tC.desc, func(t *testing.T) {
+			w := httptest.NewRecorder()
+			r := httptest.NewRequest("GET", "/", nil)
+			r.Header.Add("Authorization", "Bearer bearer")
+			authorize, b := tC.auth.Authorize(r, w)
+
+			if !b || w.Code != http.StatusOK || authorize == nil {
+				t.Errorf("Expected succesfull Authentication, but was not succesfull with code %d and body %v", w.Code, w.Body)
```
Teelevision: `w.Body` is of type `*bytes.Buffer`, so the error message will not be helpful.

http/oauth2/oauth2_test.go
```diff
@@ -80,6 +81,143 @@ func TestHandlerIntrospectError(t *testing.T) {
...
+				return
+			}
+			to, _ := security.GetTokenFromContext(authorize)
+			tok, ok := to.(*token)
+
+			if !ok || tok.value != "bearer" || tok.scope != Scope("ABC DHHG kjdk") || tok.clientID != "ClientId" || tok.userID != "UserId" {
```
Teelevision: If you were to test all these values, you should restructure the test cases. For example define the scope, client id and user id as attributes of the test case and then create the authenticator inside the test (like you did in `TestAuthenticationWithErrors`). Because if I would add a test case that did not exactly have these scope, client id and user id, it would not work.

http/oauth2/oauth2_test.go
```diff
@@ -80,6 +81,143 @@ func TestHandlerIntrospectError(t *testing.T) {
...
+	_, b := auth.Authorize(r, w)
+
+	resp := w.Result()
+	body, err := ioutil.ReadAll(resp.Body)
+	defer resp.Body.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if b {
+		t.Errorf("Expected error in Authentication, but was succesfull with code %d and body %v", resp.StatusCode, string(body))
```
Teelevision: In the success case the response would not have any code or body. I elaborated this in another comment.

http/oauth2/oauth2_test.go
```diff
@@ -80,6 +81,143 @@ func TestHandlerIntrospectError(t *testing.T) {
...
+	}
+	if got, ex := w.Code, http.StatusForbidden; got != ex {
+		t.Errorf("Expected status code %d, got %d", ex, got)
+	}
+	if got, ex := string(body), "Forbidden - requires scope \"DE\"\n"; got != ex {
```
Teelevision: Better not exactly check the body, as this might change is is not really important compared to the correct status code of the response.

http/oauth2/oauth2_test.go
```diff
@@ -80,6 +81,143 @@ func TestHandlerIntrospectError(t *testing.T) {
...
+		t.Errorf("Expected status code %q, got %q", ex, got)
+	}
+}
+
+func TestAuthenticationWithErrors(t *testing.T) {
+	testCases := []struct {
+		desc         string
+		returnedErr  error
+		expectedCode int
+		expectedBody string
```
Teelevision: Would not check body, as it might change and the code is already enough. The body is more for humans to understand the problem.

http/oauth2/oauth2_test.go
```diff
@@ -80,6 +81,143 @@ func TestHandlerIntrospectError(t *testing.T) {
...
+			_, b := auth.Authorize(r, w)
+
+			resp := w.Result()
+			body, err := ioutil.ReadAll(resp.Body)
+			defer resp.Body.Close()
+			if err != nil {
+				t.Fatal(err)
+			}
+			if b {
+				t.Errorf("Expected error in Authentication, but was succesfull with code %d and body %v", resp.StatusCode, string(body))
```
Teelevision: No code and body if it was successful. Also typo:
```suggestion
				t.Errorf("Expected error in authentication, but was succesful with code %d and body %v", resp.StatusCode, string(body))
```

http/security/apikey/authorizer_test.go
```diff
@@ -0,0 +1,87 @@
...
+	_, b := auth.Authorize(r, w)
+
+	resp := w.Result()
+	body, err := ioutil.ReadAll(resp.Body)
+	defer resp.Body.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if b {
+		t.Errorf("Expected error in Authentication, but was succesfull with code %d and body %v", resp.StatusCode, string(body))
```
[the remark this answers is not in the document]
SchlottiP: The default value for the status code is 200. I would expect that the body is empty and the status code is 200 but I tried to add all information to the error that are present and might be helpful

http/jsonapi/generator/internal/articles/open-api_test.go
```diff
@@ -308,16 +311,24 @@ type Service interface {
...
-Router implements: Articles Test Service
+RouterWithAuthentication implements: Articles Test Service
 
 Articles Test Service
 */
-func Router(service Service) *mux.Router {
+func RouterWithAuthentication(service Service, authBackend AuthorizationBackend) *mux.Router {
+	if authBackend != nil {
+		authBackend.Init()
+	}
```
Teelevision: If you remove the `Router()` method as I suggested below, you can simply assume that `authBackend` is never `nil`.

http/jsonapi/generator/internal/pay/open-api_test.go
```diff
@@ -110,9 +132,20 @@ func GetPaymentMethodsHandler(service Service) http.Handler {
 CreatePaymentMethodSEPAHandler handles request/response marshaling and validation for
  Post /beta/payment-methods/sepa-direct-debit
 */
-func CreatePaymentMethodSEPAHandler(service Service) http.Handler {
+func CreatePaymentMethodSEPAHandler(service Service, authBackend AuthorizationBackend) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		defer errors.HandleRequest("CreatePaymentMethodSEPAHandler", w, r)
+		// Only do Authentication if a authentication Backend is available. Otherwise this is handled somewhere else
+		if authBackend != nil {
```
Teelevision: I think `authBackend` can be assumed to never be `nil`, if the `Router()` function is removed as I suggested.

http/jsonapi/generator/internal/articles/open-api_test.go
```diff
@@ -45,12 +45,15 @@ type MapTypeNumber map[string]float32
 
 // MapTypeString ...
 type MapTypeString map[string]string
+type AuthorizationBackend interface {
+	Init()
+}
```
Teelevision: If there is no authentication this doesn't serve any purpose. So I'll suggest to remove `AuthorizationBackend` and `RouterWithAuthentication()` and keep `Router()` as it was.

http/jsonapi/generator/generate_security.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/pace/bricks/maintenance/errors"
+)
+
+const (
+	authBackendInterface = "AuthorizationBackend"
+	authFuncPrefix       = "Authorize"
+)
+
+// buildSecurityBackendInterface builds the interface that is used to do the authentication.
+// It creates one method for each security type and a Init-Method for handling the securityConfigs
```
Teelevision:
```suggestion
// It creates one method for each security type and an init method for handling the securityConfigs.
```

http/jsonapi/generator/generate_security.go
```diff
@@ -0,0 +1,117 @@
...
+// the Methods are named AutenticateNAME and Init
```
Teelevision:
```suggestion
// The Methods are named AuthenticateNAME and Init.
```

http/jsonapi/generator/generate_security.go
```diff
@@ -0,0 +1,117 @@
...
+func (g *Generator) buildSecurityBackendInterface(schema *openapi3.Swagger) error {
+	securitySchemes := schema.Components.SecuritySchemes
+	//r contains the methods for the security interface
```
Teelevision:
```suggestion
	// r contains the methods for the security interface
```

http/jsonapi/generator/generate_security.go
```diff
@@ -0,0 +1,117 @@
...
+	r := &jen.Group{}
+	//configs contains the names and types of the needed configs for the init method
```
Teelevision:
```suggestion
	// configs contains the names and types of the needed configs for the init method
```

http/jsonapi/generator/generate_security.go
```diff
@@ -0,0 +1,117 @@
...
+	// (that initializes the Backend with the security configs)
```
Teelevision:
```suggestion
	// (that initializes the backend with the security configs)
```

http/jsonapi/generator/generate_test.go
```diff
@@ -34,7 +36,15 @@ func TestGenerator(t *testing.T) {
 				t.Fatal(err)
 			}
 
+			if updatedExpectedFiles {
```
Teelevision: `updatedExpectedFiles` is always false. This looks like you did this for debugging reasons and forgot to change it back.

http/oauth2/oauth2.go
```diff
@@ -8,33 +8,35 @@ package oauth2
 
 import (
 	"context"
+	"fmt"
 	"net/http"
-	"strings"
 
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
```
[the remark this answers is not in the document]
Teelevision: Still open

http/oauth2/middleware/scopes_middleware_test.go
```diff
@@ -85,11 +85,11 @@ func setupRequest() *http.Request {
 	return req
 }
 
-type tokenIntrospecter struct {
+type TokenIntrospector struct {
```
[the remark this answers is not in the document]
Teelevision: Still open

http/oauth2/oauth2.go
```diff
@@ -8,33 +8,35 @@ package oauth2
...
+}
+
+// GetValue returns the oauth2 token for the current user, the value is part of the answer of the introspection
+func (t *token) GetValue() string {
+	return t.value
+}
+
+// Deprecated: NewMiddleware creates a new Oauth middleware - Deprecated for generated apis,
+// use the generated AuthenticationBackend of the API with oauth2.Authorizer
 func NewMiddleware(backend TokenIntrospecter) *Middleware {
```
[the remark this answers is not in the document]
Teelevision: Oh, and put the `func (m *Middleware) Handler` right after `func NewMiddleware`.

http/oauth2/oauth2.go
```diff
@@ -43,49 +45,63 @@ func NewMiddleware(backend TokenIntrospecter) *Middleware {
...
-		log.Req(r).Info().
-			Str("client_id", t.clientID).
-			Str("user_id", t.userID).
-			Msg("Oauth2")
-
-		span.LogFields(olog.String("client_id", t.clientID), olog.String("user_id", t.userID))
-
-		next.ServeHTTP(w, r.WithContext(ctx))
-	})
+func validateScope(ctx context.Context, w http.ResponseWriter, req Scope) bool {
```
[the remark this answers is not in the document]
Teelevision: Still open.

http/oauth2/oauth2.go
```diff
@@ -84,7 +84,12 @@ func introspectRequest(r *http.Request, w http.ResponseWriter, tokenIntro TokenI
 			log.Req(r).Info().Msg(err.Error())
 			http.Error(w, err.Error(), http.StatusUnauthorized)
 			return nil, false
```
Teelevision: `case ErrInvalidToken` is identical to the `default` case, so you can omit it. Or explicitly add `fallthrough`

Teelevision:
Oh, and please vendor your dependencies: `go mod vendor` and then `go mod tidy`.

Teelevision:
Also we should squash this when merging.

http/oauth2/authorizer.go
```diff
@@ -0,0 +1,62 @@
...
+	Description       string
+	Implicit          *Flow
+	Password          *Flow
+	ClientCredentials *Flow
+	AuthorizationCode *Flow
+}
+
+// NewAuthenticator creates an Authorizer for a specific TokenIntrospecter
+// This Authorizer does not check the scope
+func NewAuthenticator(introspector TokenIntrospecter, cfg *Config) *Authorizer {
```
Teelevision: Should be called `NewAuthorizer`.

PR #117, opened 2019-11-15
#33: added /health/all endpoint for all services

SchlottiP:
Added /health/all endpoint that lists the result of all registered health checks.


maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -68,13 +74,55 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// Does all HealthChecks and lists the result in the response body
+func checkAll(w http.ResponseWriter) {
+	results := make(map[string]error)
+	// do all checks and save the results
+	checks.Range(func(key, value interface{}) bool {
```
Ferlonas: As far as I can see, this is basically a generalized implementation of the singular healthCheck.
In both cases you use (name string, hc HealthCheck) in order to check for initialization errors and health status.
You could extract this loop and move it to a separate function, and treat singular healthChecks as a map with a size of 1.
That would eliminate the need for duplicate code.

Teelevision: At least you should separate the health check logic from the response generation.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -68,13 +74,55 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		if val, isIn := initErrors.Load(name); isIn {
+			err := val.(error)
+			results[name] = err
+			return true
+		}
+		_, res := hc.HealthCheck()
+		results[name] = res
+		return true
+	})
+	//create the response from the HealthCheck results
```
Teelevision:
```suggestion
	// create the response from the HealthCheck results
```

maintenance/health/servicehealthcheck/healthchecker_test.go
```diff
@@ -48,30 +49,75 @@ func setup(t *testHealthChecker) *http.Response {
...
-	resp := setup(&testHealthChecker{name: "TestHandlerHealthCheckErr", healthCheckErr: true})
+	resp := setupOneHealthCheck(&testHealthChecker{name: "TestHandlerHealthCheckErr", healthCheckErr: true})
 	if resp.StatusCode != 503 {
 		t.Errorf("Expected /health to respond with 503, got: %d", resp.StatusCode)
 	}
 	helperCheckResponse(t, "ERR", resp)
 
 }
 
+func TestHealthCheckAllWithErrors(t *testing.T) {
```
Teelevision: Please add a test for the successful case as well.

maintenance/health/servicehealthcheck/healthchecker_test.go
```diff
@@ -48,30 +52,97 @@ func setup(t *testHealthChecker) *http.Response {
...
+	if resp.StatusCode != statusCode {
+		t.Errorf("Expected /health to respond with %d, got: %d", statusCode, resp.StatusCode)
+	}
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Fatal(err)
+	}
+	checkRes := strings.Split(string(data), "\n ")
+	if exp := len(checkRes); exp != 4 {
+		t.Errorf("Expected 3 HealthCheck results, but got %v", len(hcs)-1)
```
Ferlonas: It's probably more future-proof to use len(expected) as count here instead of 3.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -41,23 +43,19 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	name := splitRoute[1]
-	hcInterface, isIn := checks.Load(name)
-	if !isIn {
-		h.writeError(w, errors.New("health check not registered"), http.StatusNotFound, name)
+	if name == "all" {
```
Teelevision: Please also add a check to the `RegisterHealthCheck` that prevents registering a check with the name `"all"`.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -68,13 +66,55 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		err := val.(error)
+		return err, false
+	}
+	// this is the actual health check
+	healthy, err := hc.HealthCheck()
+	return err, healthy
+}
+
+// Does all HealthChecks and lists the result in the response body
+func checkAll(w http.ResponseWriter) {
```
Teelevision: You have a function `check()` and `checkAll()`, which look very similar, but ultimately behave totally different. What bothers me the most is that the one is writing the response directly and the other does not. And I think @Ferlonas' idea to generalise checking one/many was a good idea. And it should make it easy to separate checking and response generation. Like `ServeHTTP()` does all the response generation and it looks somewhat like this:
```go
checks := allChecks
if name != "all" {
	checks = map[string]HealthCheck{name: allChecks[name]} // but obviously check that allChecks[name] is defined
}
result := check(checks) // func check(map[string]HealthCheck) map[string]error
if name != "all" {
	// write result[name]
} else {
	for name, hc := range result {...}
}
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -68,13 +66,55 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+}
+
 func (h *handler) writeError(w http.ResponseWriter, err error, errorCode int, name string) {
 	w.Header().Set("Content-Type", "text/plain")
 	w.WriteHeader(errorCode)
 	if _, err := w.Write([]byte("ERR")); err != nil {
 		log.Warnf("could not write output: %s", err)
 	}
-	log.Warnf("healthcheck %q was not healthy: %v", name, err)
+	log.Warnf("healthCheck %q was not healthy: %v", name, err)
```
Teelevision:
```suggestion
	log.Warnf("health check %q was not healthy: %v", name, err)
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -41,31 +45,68 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+}
 
-	// this is the actual health check
-	if healthy, err := hc.HealthCheck(); healthy {
-		w.Header().Set("Content-Type", "text/plain")
-		w.WriteHeader(http.StatusOK)
-		if _, err := w.Write([]byte("OK\n")); err != nil {
-			log.Warnf("could not write output: %s", err)
+// Checks one health check
```
Teelevision:
```suggestion
// Checks one or many health checks
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -41,31 +45,68 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		return
+	hcs := &checks
+	checkAll := name == nameAllHealthCheck
+	if !checkAll {
+		hcInterface, isIn := checks.Load(name)
+		if !isIn {
+			h.writeError(w, errors.New("health check not registered"), http.StatusNotFound, name)
+			return
+		}
+		hc := hcInterface.(HealthCheck)
```
Teelevision: It's not necessary to check the type here. The value will not lose its type if you don't.

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -41,31 +45,68 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		return
+	result := check(hcs)
+	// Write the Result to the body and set the content type
+	w.Header().Set("Content-Type", "text/plain")
+	status := http.StatusOK
+	body := ""
+	for name, err := range result {
+		if checkAll {
+			if err == nil {
+				body += fmt.Sprintf("%s :OK\n ", name)
```
Teelevision:
```suggestion
				body += fmt.Sprintf("%s: OK\n", name)
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -41,31 +45,68 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			} else {
+				body += fmt.Sprintf("%s :ERR\n ", name)
```
Teelevision:
```suggestion
				body += fmt.Sprintf("%s: ERR\n", name)
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -41,31 +45,68 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	hc := hcInterface.(HealthCheck)
-
-	// If it was not possible to initialise this health check, then show the initialisation error
-	if val, isIn := initErrors.Load(name); isIn {
-		err := val.(error)
-
-		h.writeError(w, err, http.StatusServiceUnavailable, name)
-		return
+	result := check(hcs)
+	// Write the Result to the body and set the content type
```
Teelevision:
```suggestion
	// Write the result to the body and set the content type
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -74,15 +115,19 @@ func (h *handler) writeError(w http.ResponseWriter, err error, errorCode int, na
...
-	log.Warnf("healthcheck %q was not healthy: %v", name, err)
+	log.Warnf("health check %q was not healthy: %v", name, err)
 }
 
 // RegisterHealthCheck registers a HealthCheck that need to be routed. The name
 // must be unique. If the health check satisfies the Initialisable interface, it
 // is initialised before it is added.
 func RegisterHealthCheck(hc HealthCheck, name string) {
+	if name == nameAllHealthCheck {
+		log.Debugf("tried to register health check with reserved name %q", name)
```
Teelevision:
```suggestion
		log.Warnf("tried to register health check with reserved name %q", name)
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -74,15 +115,19 @@ func (h *handler) writeError(w http.ResponseWriter, err error, errorCode int, na
...
+		return
+	}
 	if _, ok := checks.Load(name); ok {
-		log.Debugf("tried to register health check with name %v twice", name)
+		log.Debugf("tried to register health check with name %q twice", name)
```
Teelevision:
```suggestion
		log.Warnf("tried to register health check with name %q twice", name)
```

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -41,31 +45,68 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	status := http.StatusOK
+	body := ""
+	for name, err := range result {
+		if checkAll {
+			if err == nil {
+				body += fmt.Sprintf("%s :OK\n ", name)
+			} else {
+				body += fmt.Sprintf("%s :ERR\n ", name)
+				status = http.StatusServiceUnavailable
+				log.Warnf("healthCheck %q was not healthy: %v", name, err)
```
Teelevision:
```suggestion
				log.Warnf("health check %q was not healthy: %v", name, err)
```

maintenance/health/servicehealthcheck/README.md
```diff
@@ -1,7 +1,24 @@
-# Health Checks for Services 
-* Makes it possible to have a health check for different services (e.g. postgres, redis)
+# Health Check
+* Makes it possible to add a health check for a service (e.g. postgres) to the list of checks checked when for `/health` and `/health/check`
```
daemonfire300: (minor) Maybe split this into two sentences.

maintenance/health/servicehealthcheck/README.md
```diff
@@ -1,7 +1,24 @@
...
 
-* A service has to implement the `HealthCheck` interface and register the health check with `servicehealthchecker.RegisterHealthCheck(hc HealthCheck, name string)`
+* `/health` returns OK(200) if all registered and required health checks are healthy and ERR(503) if any health check is not healthy
```
daemonfire300: (optional) indicate that we are talking about http status codes and not a literal response body with `OK(200)` as content. Ignore this comment if the actual contents of the response are just that.

maintenance/health/servicehealthcheck/README.md
```diff
@@ -1,7 +1,24 @@
...
 
-* `healthchecker` does NOT cache the results but offers a struct (`ConnectionState`) for caching 
+* `/health/check` returns a table with the results of all registered health checks
 
+* Each check has to implement the `HealthCheck` interface and has to be registered
+    * health checks are registered as required or optional 
+    * `servicehealthcheck.RegisterHealthCheck(hc HealthCheck, name string)` registers a health check with a unique name as required check
+    * `servicehealthcheck.RegisterOptionalHealthCheck(hc HealthCheck, name string)` registers a health check with a unique name as optional check
+    * a optional check is not checked when `/health` is called
```
daemonfire300: `an` instead of `a`

maintenance/health/servicehealthcheck/README.md
```diff
@@ -1,7 +1,24 @@
...
+    * a required check is always checked
+    
+* To make a optional health check required (or a required health check optional) the check has to be removed and registered again. This should't happen very often.
```
daemonfire300: 1. `an` instead of `a`. 
2. What does the last sentence actually mean, or what is the motivation for putting it there? That the user should not switch checks too often between required and optional? Is that even a thing that someone would da?

maintenance/health/servicehealthcheck/README.md
```diff
@@ -1,7 +1,24 @@
...
+Remove a health check with `servicehealthcheck.RemoveHealthCheck(name)`
```
daemonfire300: (minor) Is this intentionally missing a bullet point?

PR #121, opened 2019-11-20
add Grafana dashboard

Teelevision:
Add a template for a Grafana dashboard that uses the Prometheus metrics.

### DIFF
--- a/grafana-dashboard.json
+++ b/grafana-dashboard.json
@@ -1,0 +1,4440 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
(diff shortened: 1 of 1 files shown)


Teelevision:
@threez wants some graph for the number of rows affected/returned (Postgres)

Teelevision:
@threez added all metrics

threez:
@Teelevision hey, we need to change the datasource to be a variable as well, see https://www.robustperception.io/switching-between-prometheus-servers-in-grafana-using-data-source-variables

threez:
We should have more detailed graphs that explain our api requests better. We should not use `Hz` for requests counts as this is a very unusual way of talking about the data.

Teelevision:
> @Teelevision hey, we need to change the datasource to be a variable as well, see https://www.robustperception.io/switching-between-prometheus-servers-in-grafana-using-data-source-variables

Done. Also added variable for the database. And made the namespace and job variables configurable.

> We should have more detailed graphs that explain our api requests better. We should not use `Hz` for requests counts as this is a very unusual way of talking about the data.

Updated existing graphs. Added metrics for object storage and circuit breakers:
image

Teelevision:
@threez need any more graphs?

Teelevision:
@threez this now includes the latest changes

threez:
Awesome job, thx.

PR #125, opened 2019-11-22
Postgres pool metrics

### LINKED ISSUE #122: Monitoring of go pg connection pool
### Problem

Postgres connections are expensive, having many causes issues. One can deploy pgbouncer/pgpool to defer the issue, but understanding the connections caused by the application remains an important issue.

### Solution

Having a go-routine that checks the pg pool stats and exposes them via metrics.

See https://github.com/go-pg/pg/blob/master/internal/pool/pool.go#L26

### Tasks

* [ ] Document the new metrics
* [ ] Expose all of the available metrics
* [ ] Implement watcher

Teelevision:
This adds 6 new metrics about a Postgres connection pool that are collected every minute.

This pr also changes the Dockerfile and Makefile so that we use the vendored libs during test and build, documents all Postgres metrics that were previously undocumented, and fixes an unrelated linting error in `maintenance/errors/error.go`.

Closes #122


backend/postgres/metrics_test.go
```diff
@@ -0,0 +1,81 @@
...
+	case <-whenDone:
+	case <-time.After(time.Second):
+		t.Fatal("metrics were not collected after 1s")
+	}
+	// query metrics
+	resp := httptest.NewRecorder()
+	handler := promhttp.HandlerFor(metricsRegistry, promhttp.HandlerOpts{})
+	handler.ServeHTTP(resp, httptest.NewRequest("GET", "/metrics", nil))
+	body := resp.Body.String()
+	assert.Regexp(t, `pace_postgres_connection_pool_hits.*?\Wpool="test"\W`, body)
```
SchlottiP: Why do you only check, that the metric is present and not the value?

threez: We basically trust the prometheus library, we started to do the tests in order to see if the metric was collected at all.

PR #129, opened 2019-12-06
refactoring of health checks 

SchlottiP:
- "/health/" returns "OK" and 200, if all required services are healthy, "ERR" and "503" if not

- "/health/check" returns two lists. One for the required and one for the optional checks, the list shows the name, the state and a description of the state for each registered health check.

- "/health/liveness" returns always "OK" and 200

- "/health/readiness" is configureable with `healthReadinessCheck(func(http.ResponseWriter, *http.Request)) `, default: returns "OK" and 200


maintenance/health/servicehealthcheck/readable_health_handler.go
```diff
@@ -0,0 +1,40 @@
...
+package servicehealthcheck
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+)
+
+// saves length of the longest name for the column width in the table. 20 characters width is the default
+var longestCheckName = 20
```
daemonfire300: If this a package internal value and we have no way of configuring (did I miss something) then we can make this a `const`, can't we?

maintenance/health/servicehealthcheck/readable_health_handler.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (h *readableHealthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+
+	reqChecks := check(&requiredChecks)
+	optChecks := check(&optionalChecks)
+
+	status := http.StatusOK
+	table := "%-" + strconv.Itoa(longestCheckName) + "s   %-3s   %s\n"
+	body := "Required Services: \n"
+	for name, res := range reqChecks {
```
daemonfire300: Suggestion: https://golang.org/pkg/strings/#Builder

maintenance/health/servicehealthcheck/readable_health_handler.go
SchlottiP: The Value is changed for each new registered health check (in `RegisterHealthCheck` and `RegisterOptionalHealthCheck`)

daemonfire300: Sorry, I must have misread the comment. It's explained right there :D

maintenance/health/servicehealthcheck/healthchecker.go
```diff
@@ -4,85 +4,101 @@
...
-		h.writeError(w, errors.New("health check not registered"), http.StatusNotFound, name)
-		return
-	}
-	hc := hcInterface.(HealthCheck)
-
-	// If it was not possible to initialise this health check, then show the initialisation error
-	if val, isIn := initErrors.Load(name); isIn {
-		err := val.(error)
+const (
+	// Err State of a service, if a error occurred during the health check of the service
```
daemonfire300: `an error`

PR #130, opened 2019-12-10
Fixes Tests

SchlottiP:
Fixes Tests that failed on master


http/oauth2/oauth2_test.go
```diff
@@ -197,25 +197,25 @@ func TestAuthenticationWithErrors(t *testing.T) {
 		{
 			desc:         "token introspecter returns ErrBadUpstreamResponse",
 			returnedErr:  ErrBadUpstreamResponse,
-			expectedCode: 502,
+			expectedCode:  http.StatusBadGateway,
```
daemonfire300: `gofmt` both files please --> I think there is an unintentional white-space

daemonfire300: `gofmt` both files please --> I think there is an unintentional white-space

PR #134, opened 2019-12-17
don't log the health requests (#131)

SchlottiP:
added middleware that applies another middleware only if the request path not matches with any prefix of a list of ignored prefixes. Used the new Middleware for the logging and tracing of requests


maintenance/log/handler.go
```diff
@@ -9,23 +9,27 @@ import (
...
 
+	"github.com/pace/bricks/maintenance/util"
 	"github.com/rs/zerolog/hlog"
 	"github.com/rs/zerolog/log"
 )
 
 // Handler is a middleware that handles all of the logging aspects of
 // any incoming http request
-func Handler() func(http.Handler) http.Handler {
+func Handler(ignoredPrefixes ...string) func(http.Handler) http.Handler {
```
Teelevision: In order to maintain backwards compatibility, even if we in the future decide to add other options as well, please use [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) for this.

maintenance/tracing/tracing_test.go
```diff
@@ -10,30 +10,25 @@ import (
...
-func TestHandlerIgnore(t *testing.T) {
-	r := mux.NewRouter()
-	r.Use(Handler("/test"))
-	r.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {})
-
-	rec := httptest.NewRecorder()
-	req := httptest.NewRequest("GET", "/test", nil)
-
-	r.ServeHTTP(rec, req)
-}
```
Teelevision: Is there any reason why you removed this test? I understand that there is no explicit validation in this test, but at least it will show, that the tracing does not panic. Maybe modify the test like you did with `TestHandler`.

maintenance/util/ignore_prefix_middleware.go
```diff
@@ -0,0 +1,35 @@
...
+
+package util
+
+import (
+	"net/http"
+	"strings"
+)
+
+// IgnorePrefixMiddleware is a wrapper for another middleware.
+// It only calls the actual middleware if non of the ignoredPrefixes is prefix of the request path
```
Teelevision:
```suggestion
// It only calls the actual middleware if none of the ignoredPrefixes is prefix of the request path
```

maintenance/tracing/tracing.go
```diff
@@ -79,11 +68,11 @@ func (h *traceHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // Handler generates a tracing handler that decodes the current trace from the wire.
 // The tracing handler will not start traces for the list of ignoredPrefixes.
 func Handler(ignoredPrefixes ...string) func(http.Handler) http.Handler {
```
Teelevision: Same as above: please use functional options.

maintenance/log/handler.go
Teelevision: This is still open.

maintenance/util/ignore_prefix_middleware.go
```diff
@@ -0,0 +1,54 @@
...
+func WithoutPrefixes(prefix ...string) ConfigurableMiddlewareOption {
+	return func(mdw *ConfigurableMiddleware) error {
+		mdw.ignoredPrefixes = append(mdw.ignoredPrefixes, prefix...)
+		return nil
+	}
+}
+
+// NewIgnorePrefixMiddleware creates a ConfigurableMiddleware
+// actualMiddleware is the actual middleware, that is called if the request is not ignored
+func NewIgnorePrefixMiddleware(next, actualMiddleware http.Handler, cfgs ...ConfigurableMiddlewareOption) *ConfigurableMiddleware {
```
Teelevision: As with `ConfigurableMiddleware`: the naming should not be "middleware" as it returns a handler. As well as `actualMiddleware`.

maintenance/tracing/tracing.go
```diff
@@ -81,11 +69,11 @@ func (h *traceHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // Handler generates a tracing handler that decodes the current trace from the wire.
 // The tracing handler will not start traces for the list of ignoredPrefixes.
-func Handler(ignoredPrefixes ...string) func(http.Handler) http.Handler {
+func Handler() func(http.Handler) http.Handler {
```
Teelevision: Should be backwards compatible. I think it would be ok to just revert this file and leave it as it is, as it was working already.

http/router.go
```diff
@@ -27,11 +28,9 @@ func Router() *mux.Router {
 	r.Use(errors.Handler())
 
 	// for logging
-	r.Use(log.Handler())
+	r.Use(util.NewIgnorePrefixMiddleware(log.Handler(), "/health"))
 
-	r.Use(tracing.Handler(
-		// no tracing for these prefixes
-		"/metrics",
+	r.Use(tracing.Handler("/metrics",
```
Teelevision: There is no actual change to this line, please revert it.

PR #135, opened 2019-12-20
Upgrade github.com/rs/zerolog from 1.8.0 to 1.17.2

Smuerdt:
With zerolog@1.9.0 the context aware logging was improved, because
subroutines started with the same context can now no longer overwrite
their logger in this shared context.

https://github.com/rs/zerolog/commit/9cd6f6eef2d7e9041629e2dd853778a9b2abddbd

Bigger example:
https://gist.github.com/Smuerdt/8d7733cdb563041c6a6404fc3cc3a912


Teelevision:
Small breaking change when using `WithContext()`, but that's ok. Code that is using bricks has to change accordingly:
```diff
-	suite.ctx = log.With().Logger().WithContext(context.Background())
+	l := log.With().Logger()
+	suite.ctx = l.WithContext(context.Background())
```

PR #146, opened 2020-01-15
Add shopspring/decimal type for monetary data to open-api generator

Smuerdt:
This change introduces the decimal type from https://github.com/shopspring/decimal in the api generator. To make use of it, a new `string` `format` is introduced. By defining a `string` as `format` `decimal` in any valid open-api spec, the code generator will use the new type `runtime.Decimal` for this attribute. `runtime.Decimal` is a thin "wrapper" around [decimal.Decimal](https://github.com/shopspring/decimal) to cope with the problem in our used jsonapi package which only allows custom types for primitive types: https://github.com/google/jsonapi/blob/d0428f63eb511f08685bda2d0a6af1d7532fb1c0/README.md#custom-types


http/jsonapi/generator/internal/pay/open-api.json
```diff
@@ -452,14 +452,36 @@
         }
       }
     },
-    "/beta/transaction": {
+    "/beta/transaction/{pathDecimal}": {
```
daemonfire300: Why would we have a decimal in the transaction path?

Smuerdt: That's just a test case for the generator and does not reflect a real api.

PR #153, opened 2020-01-23
Add a circuit breaker as ChainableRoundTripper

Smuerdt:
This pr adds a circuit breaker for connectivity/connection issues to the `ChainableRoundTripper` stack.


Smuerdt:
Work in progress, please do not merge right now.

http/transport/circuit_breaker_tripper.go
```diff
@@ -0,0 +1,89 @@
...
+	if err := prometheus.Register(stateSwitchCounterVec); xerrors.As(err, &are) {
+		stateSwitchCounterVec, ok = are.ExistingCollector.(*prometheus.CounterVec)
+		if !ok {
+			panic(fmt.Sprintf(`existing "pace_http_circuit_breaker_state_switch_total" collector no CounterVec, but %T`, are.ExistingCollector))
+		}
+	} else if err != nil {
+		panic(err)
+	}
+
+	settings.OnStateChange = func(_ string, from, to gobreaker.State) {
```
Teelevision: It should not remove existing event handler:
```go
handler = settings.OnStateChange
settings.OnStateChange = func(s string, from, to gobreaker.State) {
	if handler != nil {
		handler(s, from, to)
	}
	// ... rest ...
}
```

http/transport/circuit_breaker_tripper.go
```diff
@@ -0,0 +1,89 @@
...
+	})
+}
+
+func NewCircuitBreakerTripper(settings gobreaker.Settings) *circuitBreakerTripper {
+	if settings.Name == "" {
+		panic("name is mandatory for circuit breaker")
+	}
+
+	stateSwitchCounterVec := prometheus.NewCounterVec(prometheus.CounterOpts{
+		ConstLabels: prometheus.Labels{"name": settings.Name},
```
Teelevision: What happens if you register the second circuit breaker? Do we then have both or does it error because it exists already?

http/transport/circuit_breaker_tripper_test.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func TestCircuitBreakerTripper(t *testing.T) {
+	req := httptest.NewRequest("GET", "/foo", nil)
+
+	t.Run("with_default_settings", func(t *testing.T) {
+		breaker := NewDefaultCircuitBreakerTripper("testcircuitbreaker")
+		chain := Chain(breaker).Final(&failingRoundTripper{})
+
+		for i := 0; i < 6; i++ {
```
Teelevision: Why is it 6 instead of 5? I thought it would open after the 5th failed request, so the 6th should already be returning the open state error. What am I missing?

http/transport/circuit_breaker_tripper_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	})
+
+	t.Run("resp_object_untouched", func(t *testing.T) {
+		wantBodyStr := "some text in the body which should be unaffected by the circuit breaker"
+		breaker := NewDefaultCircuitBreakerTripper("testcircuitbreaker")
+		chain := Chain(breaker).Final(&transportWithBody{body: wantBodyStr})
+
+		resp, err := chain.RoundTrip(req)
+		assert.NoError(t, err, "expected no err, got err=%q", err)
```
Teelevision:
```suggestion
		require.NoError(t, err, "expected no err, got err=%q", err)
```

`import "github.com/stretchr/testify/require"`

`require` prints the error and aborts the test while `assert` prints the error only. So use `require` if the following conditions will inevitable fail because the current failed.

http/transport/circuit_breaker_tripper_test.go
Smuerdt: Because the rule is `counts.ConsecutiveFailures > 5` so he is opening after the 5th (with the 6th) 
failed consecutive failed request. The comment might be not clear enough, I will adapt it.

http/transport/circuit_breaker_tripper.go
Smuerdt: If the `name` is the same, the second circuit breaker will write to the same `CounterVec`. See https://github.com/pace/bricks/blob/cdade781f7055b617e75c45f3f608d7a9fc47cd1/http/transport/circuit_breaker_tripper.go#L53

PR #157, opened 2020-01-29
Skip validation for fields of map type

### LINKED ISSUE #156: Structs with fields of type map fails to validate
When using `additionalProperties` the tool will create a map for that field/struct. However this map will not be able to pass validation .

It looks like for a `map[string]interface{}` the validation tag should be `validate:"-"`: #https://github.com/asaskevich/govalidator/issues/58

agandrabur:
Fixes #156 

Checks if the field has `additionalProperties` and adds `validator:"-"` instead of `validator:"optional"` or `validator:"required"`


threez:
@agandrabur hey could you check the `http/jsonapi/generator/internal/articles/open-api.json` in the pr it looks as if the spacing was changed that caused a lot of "changed lines", please reduce to what is required for the PR ;-)

PR #163, opened 2020-02-06
Upgrade to Golang 1.13 (does not include updating packages) this is a…

daemonfire300:
_(does not include updating packages)_ This is a simple go.mod version bump

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/pace/bricks
 
-go 1.12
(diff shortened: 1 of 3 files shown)


Smuerdt:
Followup: wouldn't it be good to also fixate the go verion here: https://github.com/pace/bricks/blob/dbd3aff37f809570e65f3993b132feee3446498e/docker-compose.yml#L47

Issue #160, opened 2020-02-05
Autogenerate go pg filter/page/sort


threez:
Extract the filter/paging/sorting from the http request and enable direct filtering of go pg queries based on the jsonapi specification.

```go
var r *http.Request
filter := jsonapi.FilterFromRequest(r)
paging := jsonapi.PagingFromRequest(r)
sorting := jsonapi.SortingFromRequest(r)
var objs []model.SomeObject
sorting.Apply(paging.Apply(filter.Apply(db.Model(&objs)))).Select()
```

PR #164, opened 2020-02-11
#160: Extract the filter/paging/sorting from the http request 

SchlottiP:
resolves: https://github.com/pace/bricks/issues/160


http/jsonapi/runtime/standart_params.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return func(query Query) Query {
+		if pageNr == 0 {
+			query.Offset(0)
+		} else {
+			query.Offset(pageSize * (pageNr - 1))
+		}
+		query.Limit(pageSize)
```
threez: would be good if we can set a min,max range for the pageSize

http/jsonapi/runtime/standart_params.go
```diff
@@ -0,0 +1,117 @@
...
+		for _, val := range sorting {
+			if val == "" {
+				continue
+			}
+			order = " ASC"
+			if strings.HasPrefix(val, "-") {
+				order = " DESC"
+			}
+			val = strings.TrimPrefix(val, "-")
+			query.Order(mapper(val) + order)
```
threez: I think we need to provide an array of allowed sorting keys, otherwise we will allow SQLi (https://en.wikipedia.org/wiki/SQL_injection)

http/jsonapi/runtime/standart_params.go
```diff
@@ -0,0 +1,117 @@
...
+		for name, vals := range filter {
+			if len(vals) == 0 {
+				continue
+			}
+			var filterValues []string
+			for _, val := range vals {
+				filterValues = append(filterValues, strings.Split(val, ",")...)
+			}
+			if len(filterValues) == 1 {
+				query.Where(name+"=?", filterValues[0])
```
threez: the same issue with allowed filter values here, to prevent SQLi

http/jsonapi/runtime/standart_params.go
```diff
@@ -0,0 +1,117 @@
...
+				continue
+			}
+			query.Where(name+" IN (?)", strings.Join(filterValues, ","))
```
threez: and here (SQLi)

threez:
@SchlottiP generally looks good but we should add some protection to the query generator

SchlottiP:
* added possibility to mark a sort or filter parameter as invalid (with the mapper)
* Added max. and min. for the page size 
* Did not anything to prevent sql injection, because ORMs should do this by preparing the statement bevor replacing the `?` with values, (go-pg is SQL injection save https://github.com/go-pg/pg/wiki/Writing-Queries)

http/jsonapi/runtime/standart_params.go
```diff
@@ -4,24 +4,46 @@
...
+	"github.com/caarlos0/env"
+	"github.com/pace/bricks/maintenance/log"
 )
 
 // QueryOption is a function that applies an option (like sorting, filter or pagination) to a database query
 type QueryOption func(query Query) Query
 
 // Query based on orm.Query but only the needed function
 // guaranties testability of all of the following functions
+// An Implementation MUST prepare statements to prevent SQL injection
```
threez: We can't use prepared statements as we are using PGBouncer with transaction mase isolation, is this really required?

http/jsonapi/runtime/standart_params.go
```diff
@@ -4,24 +4,46 @@
...
+	MaxPageSize int `env:"MAX_PAGE_SIZE" envDefault:"100"`
+	MinPageSize int `env:"MIN_PAGE_SIZE" envDefault:"1"`
+}
+
+var cfg config
+
+func init() {
+	err := env.Parse(&cfg)
+	if err != nil {
+		log.Fatalf("Failed to parse server environment: %v", err)
```
threez: lets be more specific, otherwise we don't understand right from the message, the cause of the issue:

```suggestion
		log.Fatalf("Failed to parse jsonapi params from environment: %v", err)
```

http/jsonapi/runtime/standart_params.go
```diff
@@ -52,47 +78,73 @@ func PagingFromRequest(r *http.Request) (QueryOption, error) {
...
 }
 
 // FilterFromRequest adds filter to a query based on the request query parameter
 // filter[name]=val1,val2 results in name IN (val1, val2), filter[name]=val results in name=val
 // Database model and response type may differ, so the mapper function allows to map the name of  field from the request
-// to a database column name
+// to a database column name.
+// Will always return a QueryOptions function with all valid filters (can be a none)
+// if any filter are invalid a error with a list of all invalid filters is returned
+// Is SQLi save if the query uses prepared statements
```
threez: same on the prepared statements

http/jsonapi/runtime/standart_params.go
```diff
@@ -102,16 +154,21 @@ func FilterFromRequest(r *http.Request, mapper Mapper) (QueryOption, error) {
...
+	if len(invalidFilter) != 0 {
+		return filterQueryOption, fmt.Errorf("at least one filter parameter is not valid: %q", strings.Join(invalidFilter, ","))
+	}
+	return filterQueryOption, nil
 }
 
 // DefaultMapper maps an query field directly name to the same name as database column name
-func DefaultMapper(in string) string {
-	return in
+func DefaultMapper(in string) (string, bool) {
```
threez: so should we have a mapper now, that is getting a list and returns true if something is in the list? maybe even a map ...

SchlottiP: a Map only allows a explicit whitelist of parameters
a function allows something like a regex to check if the fieldname is valid
=> Both an option, a map is more restrictive by default. I will change it to use a map

threez:
@SchlottiP please rebase on the current master so that the amout of changes is reduced

SchlottiP:
rebased Branch, kann you review @threez or should I ask Marius or any other person of the team?

http/jsonapi/runtime/standart_params.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	filterQueryOption := func(query Query) Query {
+		for name, filterValues := range filter {
+			if len(filterValues) == 0 {
+				continue
+			}
+
+			if len(filterValues) == 1 {
+				query.Where("?=?", name, filterValues[0])
```
threez: not sure it that works, could you add integration tests in general

SchlottiP: Still not Sure if this was okay, changed it before the test was ready

Makefile
```diff
@@ -46,3 +46,6 @@ testserver:
 
 ci:
 	go test -mod=vendor -count=1 -v -cover -race -covermode=atomic -coverprofile=coverage.out ./...
+
+testone:
```
threez, line 50: this may be a left over? rest looks good

PR #167, opened 2020-02-20
add routine package

Teelevision:


pkg/routine/routine.go
```diff
@@ -0,0 +1,79 @@
...
+	routineCtx = oauth2.ContextTransfer(ctx, routineCtx)
+	routineCtx = errors.ContextTransfer(ctx, routineCtx)
+	// apply timeout
+	routineCtx, cancel = context.WithTimeout(routineCtx, timeout)
+	// register context to be cancelled when the program is shut down
+	contextsMx.Lock()
+	contexts[routineCtx] = cancel
+	contextsMx.Unlock()
+	// deregister the above if context is done
+	go func() {
```
threez: would not start a sep. go routine but move that as a defer func into the func below.

pkg/routine/routine.go
```diff
@@ -0,0 +1,79 @@
...
+	"syscall"
+	"time"
+
+	"github.com/pace/bricks/http/oauth2"
+	"github.com/pace/bricks/maintenance/errors"
+	"github.com/pace/bricks/maintenance/log"
+)
+
+// Forever is so far in the future that it basically never ends.
+var Forever = time.Duration(1<<63 - 1)
```
threez: IN the stdlib unlimited is usually == 0

pkg/routine/routine.go
```diff
@@ -0,0 +1,79 @@
...
+// Panics thrown in the function are logged and sent to sentry. The routines
+// context is cancelled if the program receives a shutdown signal (SIGINT,
+// SIGTERM).
+func Run(ctx context.Context, timeout time.Duration, routine func(context.Context)) (cancel context.CancelFunc) {
+	// transfer logger, authentication and error info
+	routineCtx := log.Ctx(ctx).WithContext(context.Background())
+	routineCtx = oauth2.ContextTransfer(ctx, routineCtx)
+	routineCtx = errors.ContextTransfer(ctx, routineCtx)
+	// apply timeout
+	routineCtx, cancel = context.WithTimeout(routineCtx, timeout)
```
threez: only add if timeout is not `0`.

pkg/routine/routine.go
```diff
@@ -0,0 +1,79 @@
...
+var Forever = time.Duration(1<<63 - 1)
+
+// Run runs the given function in a new context with the given timeout. The new
+// context inherits the logger and oauth2 authentication of the passed context.
+// Panics thrown in the function are logged and sent to sentry. The routines
+// context is cancelled if the program receives a shutdown signal (SIGINT,
+// SIGTERM).
+func Run(ctx context.Context, timeout time.Duration, routine func(context.Context)) (cancel context.CancelFunc) {
+	// transfer logger, authentication and error info
+	routineCtx := log.Ctx(ctx).WithContext(context.Background())
```
threez: As an idea, if we did have an auto-inc counter we could use that to give each of the routines a unique name and add that to the logger, that way, we can see all the log statements of one routine. the id could also be used for the context map

pkg/routine/routine.go
```diff
@@ -0,0 +1,79 @@
+// Copyright © 2019 by PACE Telematics GmbH. All rights reserved.
+// Created at 2019/10/01 by Marius Neugebauer
+
+// Package routine helps in starting background tasks.
```
threez: very good idea.

pkg/routine/routine.go
Teelevision: If it was a defer in the function below, then it would not be triggered when the context times out, but when the routine finishes.

pkg/routine/routine.go
Teelevision: Will do.

pkg/routine/routine.go
Teelevision: Will do.

pkg/routine/routine.go
Teelevision: I like the idea. Would definitely make the logs more readable. Will do. But I see no use in using the id in the `contexts` map. It's already a pointer and therefore cheap as an index.

Teelevision:
@threez and I discussed this in person, that I'll split the 3 things that this package does: transfer context, cancel context on shutdown, and applying a timeout. Applying a timeout will be left to the caller. Cancel on shutdown will be default behaviour for Run. And transfer context will be a separate method that can be called if needed.

But upon revisiting this issue it occurred to me that cancelling the context on shutdown only makes sense if it is running in the background. The only use case I see for using the parent context on the other hand is when we do something in parallel. In that case we don't want to cancel the routines on shutdown, because the caller is probably waiting for them to finish processing. The caller should decide whether to wait for them or cancel them using the returned CancelFunc. So in that case the caller would be advised to use just goroutines. Leaving us with background tasks as the only real use case for this package. Therefore I'm staying with having only the Run function.

I'm dropping the timeout though, because it is nearly as simple to apply it in the routine, and much more expressive. Before
```go
routine.Run(ctx, time.Hour, func(context.Context) {
    // logic
})
```
After:
```go
routine.Run(ctx, func(context.Context) {
    ctx, cancel := context.WithTimeout(ctx, time.Hour)
    // logic
    cancel()
})
```

PR #169, opened 2020-02-25
Add log sink to be able to send breadcrumbs to sentry

Smuerdt:
This pr introduces a log sink to be able to store logs inside a context and send them to sentry as breadcrumbs if an error happens.

Note: since https://github.com/getsentry/raven-go is no longer supported and does not support breadcrumbs, this pr moves the package to `maintenance/errors/raven` and  adds the missing breadcrumb feature.


maintenance/log/sink.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/hlog"
+	"github.com/rs/zerolog/log"
+)
+
+type sinkKey struct{}
+
+// WithSink wraps the given context in a new context with
+// the given Sink stored as value.
+func WithSink(ctx context.Context, sink *Sink) context.Context {
```
Teelevision: I think a better naming would be `ContextWithSink`. Because `log.WithSink` sounds more like you are modifying a logger.

maintenance/log/sink.go
```diff
@@ -0,0 +1,98 @@
...
+	return context.WithValue(ctx, sinkKey{}, sink)
+}
+
+// SinkFromContext returns the Sink of the given context if it exits
```
Teelevision:
```suggestion
// SinkFromContext returns the Sink of the given context if it exists
```

maintenance/log/sink.go
```diff
@@ -0,0 +1,98 @@
...
+func (s *Sink) Write(b []byte) (int, error) {
+	s.rwmutex.Lock()
+	if s.output == nil {
+		s.output = logOutput
+	}
+
+	s.logs = append(s.logs, string(b))
+	s.rwmutex.Unlock()
+
+	return s.output.Write(b) // what if n < len(b)
```
Teelevision: Confusing comment.

maintenance/errors/error.go
```diff
@@ -219,5 +219,10 @@ func (e sentryEvent) build() *raven.Packet {
 	// from env
 	packet.Extra["microservice"] = os.Getenv("JAEGER_SERVICE_NAME")
 
+	// add logs
+	if sink, ok := log.SinkFromContext(ctx); ok {
+		packet.Extra["logs"] = sink.Logs()
```
Teelevision: How is this handled by Sentry? Because it's a slice and not a string.

maintenance/log/handler.go
```diff
@@ -18,10 +18,9 @@ import (
 // any incoming http request
 func Handler() func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
-		return hlog.NewHandler(log.Logger)(
-			hlog.AccessHandler(requestCompleted)(
-				hlog.RequestIDHandler("req_id", "Request-Id")(next)))
-
+		return hlog.AccessHandler(requestCompleted)(
+			hlog.RequestIDHandler("req_id", "Request-Id")(
+				HandlerWithLogSink(log.Logger)(next)))
```
Teelevision: I believe there maybe was a reason for the `hlog.NewHandler` to be on the top of the chain. It's now in `HandlerWithLogSink` at the bottom.

maintenance/log/sink.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return sink
+}
+
+// HandlerWithLogSink returns a mux.MiddlewareFunc which
+// adds the given logger together with a Sink to the request
+// context. All logs corresponding to the request will be
+// printed and stored in the Sink for later use.
+func HandlerWithLogSink(logger zerolog.Logger, opts ...func(*Sink)) mux.MiddlewareFunc {
```
Teelevision: I think the naming is off for most functions. We have `log.HandlerWithLogSink` that doesn't take a `Sink` argument. We have `log.CustomOutput` that's only applicable to `log.SinkWith` and `log.HandlerWithLogSink`. And it kind of looks good for `log.SinkWith(log.CustomOutput(...))`, but not for `log.HandlerWithLogSink(logger, log.CustomOutput(...))`. The naming looks like you actually wanted to do a `package sink`, but you are restricted to the `log` package.

I think using functional options right now is overkill. Actually, I think using any arguments or configuration for this is not needed for now, especially as we don't really know what we want to configure. I don't expect any real impact (performance wise, for example) from just enabling the sink in `log.Handler()` generally. Don't expose any new functions or types. We don't need any of that right now. And if we need configuration in the future, we are still able to add configuration.

maintenance/errors/error.go
Smuerdt: It's handled the same way as the `oauth2.Scopes`. Please have a look in Sentry (for example pay) to see a concrete example.

maintenance/log/handler.go
Smuerdt: Good catch, I missed that `RequestIDHandler` needs the logger coming from `hlog.NewHandler`. I will change it.

maintenance/errors/error.go
```diff
@@ -158,7 +161,8 @@ type sentryEvent struct {
 }
 
 func (e sentryEvent) Send() {
-	raven.Capture(e.build(), nil)
+	_, errCh := raven.Capture(e.build(), nil)
+	<-errCh
```
Teelevision: Debugging artifact?

maintenance/errors/error.go
```diff
@@ -219,5 +223,133 @@ func (e sentryEvent) build() *raven.Packet {
...
+	if err := json.Unmarshal(sink.ToJSON(), &data); err != nil {
+		log.Ctx(ctx).Warn().Err(err).Msg("failed to prepare sentry message")
+		return nil
+	}
+
+	result := make([]*raven.Breadcrumb, len(data))
+	for i, d := range data {
+		crumb, err := createBreadcrumb(d)
+		if err != nil {
+			log.Ctx(ctx).Warn().Err(err).Msg("failed to create sentry breadcrump")
```
Teelevision:
```suggestion
			log.Ctx(ctx).Warn().Err(err).Msg("failed to create sentry breadcrumb")
```

maintenance/errors/error_test.go
```diff
@@ -72,3 +76,144 @@ func TestStackTrace(t *testing.T) {
...
+
+		panic("Sink Test Error, IGNORE")
+	})
+	log.Handler()(Handler()(mux)).ServeHTTP(rec, req)
+
+	resp := rec.Result()
+	defer resp.Body.Close()
+
+	require.Equal(t, 500, resp.StatusCode, "wrong status code")
+}
```
Teelevision: What are you testing here? I think at least the name of the test func should give a hint. Because I'm not really seeing what this test is about and what it has to do with the sink.

Teelevision: What are you testing here? I think at least the name of the test func should give a hint. Because I'm not really seeing what this test is about and what it has to do with the sink.

Teelevision: What are you testing here? I think at least the name of the test func should give a hint. Because I'm not really seeing what this test is about and what it has to do with the sink.

Teelevision: What are you testing here? I think at least the name of the test func should give a hint. Because I'm not really seeing what this test is about and what it has to do with the sink.

Teelevision: What are you testing here? I think at least the name of the test func should give a hint. Because I'm not really seeing what this test is about and what it has to do with the sink.

maintenance/errors/error.go
Smuerdt: No, intended. This change ensures that the sentry message gets send even if the process itself get's killed.

Teelevision: Oh I see. Can you add that as a comment to the code, please?

PR #171, opened 2020-02-26
routine: keep running one instance in cluster

Teelevision:


Makefile
```diff
@@ -39,6 +39,7 @@ test:
 
 integration:
 	go test -mod=vendor -count=1 -v -cover -race -run TestIntegration ./...
+	go test -mod=vendor -count=1 -v -cover -race -run Example_clusterBackgroundTask ./pkg/routine
```
daemonfire300: Is there a specific reason why we call it an Example?

daemonfire300: Nvm, you are using the example output feature of go. I see.

Teelevision: Because it's an example how to use it.

pkg/routine/config.go
```diff
@@ -0,0 +1,22 @@
...
+package routine
+
+import (
+	"time"
+
+	"github.com/caarlos0/env"
+)
+
+type config struct {
+	RedisLockTTL time.Duration `env:"ROUTINE_REDIS_LOCK_TTL" envDefault:"1s"`
```
Smuerdt: As discussed in person: We should maybe think about the rate how often the `TTL` gets refreshed. Right now it seems to be `TTL/2` (https://github.com/pace/bricks/pull/171/files#diff-ee9982e919501ff31f7336477af2941bR30). I personally think we should increase this ration to a `TTL/5` or even `TTL/10` which would make it highly unlikely in a realistic scenario, that the current master loses his lock even though he is running fine and just couldn't refresh the lock because of network problems or something else. This would mean that we also should think about raising the general `TTL` to `5s` or `10s` to avoid golang runtime specific effects of a too fast refresh rate.

pkg/routine/routine.go
```diff
@@ -18,6 +18,67 @@ import (
...
+
+// RunNamed runs a routine like Run does. Additionally it assigns the routine a
+// name and allows using options to control how the routine is run. Routines
+// with the same name show consistent behaviour for the options, like mutual
+// exclusion, across their group. By default all callers that share the same
+// redis database are members of the same group, no matter whether they are
+// goroutines of a single process or of processes running on different hosts.
+// The default redis database is configured via the REDIS_* environment
+// variables.
+func RunNamed(parentCtx context.Context, name string, routine func(context.Context), opts ...Option) (cancel context.CancelFunc) {
```
Ferlonas: If the name defines the group (at least that's how I understood the comment), then the function could be named
"RunGrouped" or "RunInGroup", with a parameter "groupName" instead of "name"

pkg/routine/redis_lock.go
```diff
@@ -0,0 +1,101 @@
...
+
+// Try to obtain a lock. Return a sub-context of ctx that is canceled once the
+// lock is lost or ctx is done.
+func obtainLock(ctx context.Context, locker *redislock.Client, key string, ttl time.Duration) context.Context {
+	// obtain lock
+	lock, err := locker.Obtain(key, ttl, nil)
+	if err == redislock.ErrNotObtained {
+		return nil
+	} else if err != nil {
+		panic(err)
```
Ferlonas: This will crash an entire process (and potentially pod) without notifying anyone about it. It Is not obvious from the outside that a panic can occur this deep in RunNamed()

pkg/routine/routine.go
Teelevision: The name refers to the routine in the group, not the group itself. The group just consists of everyone sharing a Redis database.

pkg/routine/config.go
```diff
@@ -0,0 +1,22 @@
+// Copyright © 2020 by PACE Telematics GmbH. All rights reserved.
+// Created at 2020/02/27 by Marius Neugebauer
+
+package routine
```
Teelevision: Also suggested was changing the package name to `background` to avoid confusion between goroutines and routines. And also calling the routines tasks or similar.

We should however keep the `routine` package to not break backwards compatibility.

PR #173, opened 2020-03-04
Reduce health check logging

threez:
* adds panic handler to health cehck
* uses ctx in redis and postgres for logging
* fixes docker-compose config for sentry
* removes NoopSink
* health check func and ctx for init
* updates go version to 1.14
* adds github.com/mattn/goveralls and golang.org/x/tools/cmd/cover to vendor folder


threez:
FYI @SchlottiP please check the changes as well

PR #176, opened 2020-03-09
use github.com/pace/jsonapi instead of google for better errors #175

### LINKED ISSUE #175: json api type errors are not helpful
### Problem

The `github.com/google/jsonapi` unmarshal func does type check the request but doesn't inform about the particular issue. If the caller is using a string instead of a number, the generic error `Invalid type provided` is returned.

### Solution

The library needs to be changed to report the issue.

threez:
fixes #175


http/jsonapi/generator/internal/fueling/fueling_test.go
```diff
@@ -0,0 +1,71 @@
...
+	resp := rec.Result()
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 422 {
+		t.Errorf("expected 422 got: %d", resp.StatusCode)
+		b, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			t.Fatal(err)
+		}
+		t.Error(string(b[:]))
```
Ferlonas: Why b[:] instead of just b?

http/jsonapi/generator/internal/fueling/fueling_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if strings.Contains(string(b[:]), "can't parse content: Got value \"47.8\" expected type float32: Invalid type provided") {
```
Ferlonas: That's rather specific. Shouldn't we test for the actual expected content instead of what we don't want in there?

http/jsonapi/generator/internal/fueling/fueling_test.go
threez: will fix.

http/jsonapi/generator/internal/fueling/fueling_test.go
threez: good point.

Issue #179, opened 2020-03-19
Add log breadcrumbs to log.handleError()


Ferlonas:
Currently, breadcrumbs are only attached when an http request is aborted with an error or panic. Sometimes we want to use handleError to report a message to sentry without aborting the request.

Breadcrumbs should be added there as well.

Smuerdt:
If the `context.Context` object you are passing to handleError contains a `log.Sink` this will already work out of the box. However I see a big Problem how we easily get a `log.Sink` in every "initial" context so that you can't forget it easily.

I propose to change the `log.WithContext` func in a way that it also adds a log Sink to the newly returned context. With this solution and the `log.ContextWithSink` we could promote to use these two function if a new context with logger is needed.

Teelevision:
Adding it to `log.WithContext` does not break with backwards compatibility. Everything else still works as expected. We just get extra functionality. Right?

Smuerdt:
Yes, but you need to be aware that you still need to use `log.ContextWithSink` at some places. For example if you want to directlty attach some values to the logger before calling `WithContext` on it.

```go
l := log.Logger().With().Str("ws_id", reqID).Logger()
ctx := log.ContextWithSink(l.WithContext(context.Background()), new(log.Sink))
```

And that resembles actually the biggest concern I have. There is no uniform and automatic way to get a log sink inside a context with a logger.

Teelevision:
But if the original context had the sink then the child context has it, too, because the child context just uses the same output as the parent context, I would assume.

Teelevision:
Oh, I get what you mean. We can control `log.WithContext`, but don't control `(*Logger) With Context`.

Teelevision:
What about `log.Logger()`? Can we equip the default logger with a sink maybe?

Teelevision:
Hm, a sink in the default logger leaves us with all the other logs that we don't want as well.

Smuerdt:
I also thought about this but I'm not sure if this solves our problem, as we would than have a global sink for all logs...

Teelevision:
I think this might be a bug, not a design problem. Anyway, we need more info why the breadcrumbs weren't transmitted in that case.

PR #180, opened 2020-03-26
Transfer also log.Sink between contexts in routine pkg

Smuerdt:
Related to https://github.com/pace/bricks/issues/179

### DIFF
--- a/maintenance/log/sink.go
+++ b/maintenance/log/sink.go
@@ -29,6 +29,18 @@ func SinkFromContext(ctx context.Context) (*Sink, bool) {
 	return sink, ok
 }
 
+// SinkContextTransfer gets the sink from the sourceCtx
+// and returns a new context based on targetCtx with the
+// extracted sink. If no sink is present this is a noop
+func SinkContextTransfer(sourceCtx, targetCtx context.Context) context.Context {
+	sink, ok := SinkFromContext(sourceCtx)
+	if !ok {
+		return targetCtx
+	}
+
+	return context.WithValue(targetCtx, sinkKey{}, sink)
+}
+
 // Sink respresents a log sink which is used to store
 // logs, created with log.Ctx(ctx), inside the context
 // and use them at a later point in time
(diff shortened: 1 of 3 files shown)


Teelevision:
Please fix failed checks.

maintenance/log/sink.go
```diff
@@ -29,6 +29,18 @@ func SinkFromContext(ctx context.Context) (*Sink, bool) {
 	return sink, ok
 }
 
+// ContextTransfer gets the logger and sink from the sourceCtx
+// and returns a new context based on targetCtx and these two.
+// If there is no sink in sourceCtx, a new one is created.
+func ContextTransfer(sourceCtx, targetCtx context.Context) context.Context {
```
Teelevision: Why would you want to create a sink if there isn't one in the original context?

maintenance/log/sink.go
```diff
@@ -29,6 +29,18 @@ func SinkFromContext(ctx context.Context) (*Sink, bool) {
...
+	sink, ok := SinkFromContext(sourceCtx)
+	if !ok {
+		sink = new(Sink)
+	}
+
+	return ContextWithSink(targetCtx, sink)
+}
+
```
Teelevision: It doesn't actually transfer the logger. It only transfers the sink. It should transfer both.

Smuerdt: It does in `ContextWithSInk` but doesn't matter I changed it.

pkg/routine/routine_test.go
```diff
@@ -33,13 +32,13 @@ func TestRun_createsNewContext(t *testing.T) {
...
-	ctx := logger.WithContext(context.Background())
+	var sink log.Sink
+	logger := log.Logger()
+	ctx := log.ContextWithSink(logger.WithContext(context.Background()), &sink)
 	waitForRun(ctx, func(ctx context.Context) {
 		log.Ctx(ctx).Debug().Msg("foobar")
 	})
-	require.Contains(t, buf.String(), "foobar")
+	require.Contains(t, string(sink.ToJSON()), "foobar")
 }
```
Teelevision: Please don't re-purpose this test. This test did test that the logger is transferred. Add an additional test to test that the sink is transferred, please. The sink cannot go without the logger, but the logger can go without the sink.

Smuerdt: done

PR #182, opened 2020-04-07
testable filter/sorting/pagination

SchlottiP:


http/jsonapi/runtime/standart_params.go
```diff
@@ -63,46 +60,86 @@ func (m *MapMapper) Map(value string) (string, bool) {
 	return val, isValid
 }
 
-// PaginationFromRequest extracts pagination query parameter and  returns a function that adds the pagination to a query
-func PaginationFromRequest(r *http.Request) (QueryOption, error) {
+// UrlQueryParameters contains all information that are needed for pagination, sorting and filtering.
```
daemonfire300, line 63: either pluralize information or use "that is needed", or maybe my english is not good enough :D

http/jsonapi/runtime/standart_params.go
```diff
@@ -63,46 +60,86 @@ func (m *MapMapper) Map(value string) (string, bool) {
...
+	PageSize      int
+	Order         []string
+	Filter        map[string][]interface{}
+}
+
+// ReadURLQueryParameters reads sorting, filter and pagination from requests and return a UrlQueryParameters object,
+// even if any errors occur. The returned error combines all errors of pagination, filter and sorting.
+func ReadURLQueryParameters(r *http.Request, mapper ColumnMapper, sanitizer ValueSanitizer) (*UrlQueryParameters, error) {
+	result := &UrlQueryParameters{}
+	errPagination := result.setPagination(r)
```
daemonfire300, line 77: minor issue, `setXXX` implies that result does something to the pointer (method argument) it receives. This might be confusing when looking at these methods in the future.

Mabye `readXXX` or `extractXXX`?

http/jsonapi/runtime/standart_params.go
```diff
@@ -63,46 +60,86 @@ func (m *MapMapper) Map(value string) (string, bool) {
...
+		query.Order(val)
+	}
+	return query
+
+}
+
+func (u *UrlQueryParameters) setPagination(r *http.Request) error {
 	pageStr := r.URL.Query().Get("page[number]")
 	sizeStr := r.URL.Query().Get("page[size]")
 	if pageStr == "" || sizeStr == "" {
```
daemonfire300, line 118: Can't we have a default page size? Does the jsonapi spec require both parameter to be set?

I would suggest that we allow pagination with "only" the page number.

daemonfire300: e.g., we could use `cfg.MinPageSize` if none is set?

http/jsonapi/runtime/standart_params.go
SchlottiP: I looked it up: There is no plural of "information" :)

PR #183, opened 2020-04-13
Fix race condition in RetryRoundTripper.RoundTrip()

Smuerdt:
DISCLAIMER: I am not 100 percent sure if I understood the RetryRoundTripper completely, but I got a detected race condition in the addressed function when trying to do multiple parallel http calls with a single client. 

This pr replaces the per `RetryRoundTripper` static `retry.Transport` with a factory of said type (`func() *retry.Transport`). This ensures that every request gets its own `retry.Transport` when calling `RoundTrip()`. Before this change, concurrent `RoundTrip()` calls shared the same `retry.Transport` which is a race condition on the assignment: `l.RetryTransport.Next = transportWithAttempt(l.Transport())`

### DIFF
--- a/http/transport/retry_round_tripper.go
+++ b/http/transport/retry_round_tripper.go
@@ -12,8 +12,8 @@ import (
 
 // RetryRoundTripper implements a chainable round tripper for retrying requests
 type RetryRoundTripper struct {
-	retryTransportFactory func() *retry.Transport
-	transport             http.RoundTripper
+	retryTransport *retry.Transport
+	transport      http.RoundTripper
 }
 
 // RetryCodes retries when the status code is one of the provided list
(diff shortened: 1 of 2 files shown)


http/transport/retry_round_tripper.go
```diff
@@ -12,8 +12,8 @@ import (
 
 // RetryRoundTripper implements a chainable round tripper for retrying requests
 type RetryRoundTripper struct {
-	RetryTransport *retry.Transport
```
threez, line 15: I think the issue you discovered is because of the `DefaultRetryTransport` used by more than one client. One client should not use the `DefaultRetryTransport` concurrently, can you try another fix? E.g. remove the already deprecated `DefaultRetryTransport` and only use the function `NewDefaultRetryTransport`. It would be helpful to have a test to see where the race is originating from, or at least the trace for the race.

Smuerdt: No, as stated above the race I discovered is specifically only related to a single client using the bricks `transport.NewDefaultTransportChain()` when two go routines do requests through this client concurrently. For better understanding I added a test before applying my fix (aa41025). If you do a `make test` on this commit, it will fail due to a race condition.

Issue #184, opened 2020-04-14
OAuth2 Introspection use 1.13 error handling and not return 500 on invalid tokens


daemonfire300:
# Problem

`ErrInvalidToken` is being wrapped. The code above the callee does not use `errors.Is` but explicit checks this results in http 500.

# Soluation

Use 1.13 error handling

Ferlonas:
Closed due to accidental commit by me :-(

PR #185, opened 2020-04-14
Add 1.13 error handling to oauth2 intrespection #184

daemonfire300:
This avoids 500 responses when a user token is invalid

https://github.com/pace/bricks/issues/184


daemonfire300:
We probably also need a regression test for this.

Smuerdt:
Would be nice if you could add such a test before we merge this pr. :)

Ferlonas:
Fixed with different commit already. Closing this.

PR #190, opened 2020-04-20
Silence /health logs

Smuerdt:


http/router.go
```diff
@@ -26,7 +26,7 @@ func Router() *mux.Router {
 
 	// the logging middleware needs to be registered before the error
 	// middleware to make it possible to send panics to sentry
-	r.Use(log.Handler())
+	r.Use(log.Handler("/health"))
```
Ferlonas, line 29: Please add a comment saying that the path you add here is to be silent. At first read it looks as if this handler was specifically for /health only

PR #192, opened 2020-04-21
HTTP transport: Return specific errors in some cases:

Ferlonas:
* when the retry limit is exceeded
* when the circuit is open (i.e. host is unreachable)


Ferlonas:
There's another error that gets passed through (http2.StreamError), but as it is in the standard library, I think we can check that in each client specifically. I didn't want to make our POI or PAY clients check for a specific error from the retry package.

PR #193, opened 2020-04-21
fix index out of range panic

Teelevision:


daemonfire300:
An easy fix is an easy win. I would say this PR is a win win

PR #198, opened 2020-05-04
197 fix scan parameters spaces

threez:


http/jsonapi/runtime/parameters.go
```diff
@@ -50,6 +61,7 @@ func ScanParameters(w http.ResponseWriter, r *http.Request, parameters ...*ScanP
 		case ScanInQuery:
 			// input may not be filled and needs to be parsed from the request
 			input := r.URL.Query()[param.Name]
+			fmt.Println(param.Name)
```
Teelevision, line 64: Debug artifact?

http/jsonapi/runtime/parameters_test.go
```diff
@@ -9,6 +9,63 @@ import (
...
 )
 
+func TestScanStringParametersInQuery(t *testing.T) {
+	tests := []struct {
+		path   string
+		result string
+	}{
+		{"/?q=", ""},
+		{"/?q=Super", "Super"},
+		{"/?q=Super%20Plus", "Super Plus"},
```
Teelevision, line 19: Maybe add case for `Super Plus` without any encoding.

http/jsonapi/runtime/parameters.go
```diff
@@ -94,18 +100,27 @@ func ScanParameters(w http.ResponseWriter, r *http.Request, parameters ...*ScanP
...
 	return true
 }
+
+// Scan works like fmt.Sscan except for strings, they are directly assigned
+func Scan(str string, data interface{}) (int, error) {
+	// Don't handle plain strings with sscan but directly assign the value
+	t := reflect.TypeOf(data)
+	if t.Kind() == reflect.Ptr && t.Elem().Kind() == reflect.String {
+		strPtr, ok := data.(*string)
+		if ok {
```
Teelevision, line 119: The result of both `if`s cannot be different, right? So you could have only the inner `if` to reduce complexity.

http/jsonapi/runtime/parameters_test.go
threez: That would be invalid HTTP

PR #201, opened 2020-05-19
Jsonapi with decimal

threez:


http/jsonapi/request.go
```diff
@@ -434,6 +445,48 @@ func unmarshalAttribute(
...
+	case uint:
+		dec = decimal.NewFromInt(int64(v))
+	case uint8:
+		dec = decimal.NewFromInt(int64(v))
+	case uint16:
+		dec = decimal.NewFromInt(int64(v))
+	case uint32:
+		dec = decimal.NewFromInt(int64(v))
+	case uint64:
+		dec = decimal.NewFromInt(int64(v))
```
Smuerdt, line 477: Wouldn't it be better to format uints to string instead of casting to int64? At least uint64 could overflow here.

PR #204, opened 2020-06-04
Add error hiding func to only forward explicit and context errors

Smuerdt:


maintenance/errors/context_test.go
```diff
@@ -0,0 +1,154 @@
...
+				ctx:        context.Background(),
+				err:        fmt.Errorf("%s: %w", iAmAnError, context.Canceled),
+				exposedErr: iAmAnotherError,
+			},
+			want: fmt.Errorf("%w: %s: %s", iAmAnotherError, iAmAnError, context.Canceled),
+		},
+		{
+			name: "canceled_context_no_error_nothing_exposed",
+			args: args{
+				ctx:        canceledContext(),
```
threez, line 66: calling `canceledContext` like this makes the test slow, I prefer to store the result and use it in all cases.

Teelevision: Huh? This doesn't look like something that takes more than a ms combined. Or is there more to it?

Smuerdt: I did not want to comment on this but I'm on @Teelevision side here. With this argument we would need to replace all repeating `context.Background()` calls in all test cases.

Smuerdt: I will change it nevertheless...

PR #206, opened 2020-06-09
implements a simple long polling func

threez:


http/longpoll/longpoll.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Until executes the given function fn until duration d is passed or context is canceled
+func Until(ctx context.Context, d time.Duration, fn LongPollFunc) (ok bool, err error) {
+	return Default.LongPollUntil(ctx, d, fn)
+}
+
+// LongPollUntil executes the given function fn until duration d is passed or context is canceled
+func (c Config) LongPollUntil(ctx context.Context, d time.Duration, fn LongPollFunc) (ok bool, err error) {
+	until := time.Now()
+	if d >= c.MinWaitTime && d < c.MaxWaitTime {
```
Smuerdt, line 39: I think returning an error in case `d` is out of the min/max range would make the behaviour more clear to the caller. If I understand correctly it would just silently perform a single request if `d` is bigger than `c.MaxWaitTime` or smaller than `c.MinWaitTime`.

http/longpoll/longpoll.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"context"
+	"time"
+)
+
+// LongPollFunc should return true if long polling did
+// resolve, false otherwise. If error is returned, the
+// long polling will be canceled
+type LongPollFunc func() (bool, error)
```
daemonfire300, line 11: Do we really need the boolean here? Does something speak against a dedicated error like `ErrDidNotResolve`

daemonfire300, line 11: Do we want to add a `context.Context` to the `LongPollFunc` so that it can explicitly react to context "events" (cancelation etc.) of the "parent" context, i.e, the one given by `Until(ctx context.Context, ...)`.

The caller then could still choose to ignore the context in the function, but would have the ability to use it without explicitly passing it into the function body when the function is being defined.

Just a thought of a possibly "problematic" scenario would be, the a `LongPollFunc` gets created "somewhere" and spawns a goroutine that gets terminated by a context with deadline or something similar. If we now pass by accident or whatever reasons two different contexts into `Until` we might end up with strange behaviour such as, the long polling randomly failing after 10 seconds, although the `MinWaitTime` is like 60 seconds.
We would see the context cancellation error, yes, but we would need to track down which context actually caused the cancellation. And when the definition of a `LongPollFunc` and the call to `Until` are in different scopes this might be hard to keep track off.

http/longpoll/longpoll.go
Teelevision: Why even restrict this?

http/longpoll/longpoll.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		// no long polling
+		if d <= 0 {
+			break
+		}
+
+		// long pooling desired?
+		if !time.Now().Before(until) {
+			break
```
Teelevision, line 62: If the remaining time is smaller than `c.RetryTime` this will not break, but in the select below the loop will continue for another round after the `c.RetryTime`, even though by then until was reached.

Teelevision: I would suggest creating a second context that is used in the select below instead of `until` here.

http/longpoll/longpoll.go
threez: I like to separate the resolve from the error handling. Having the call only once is no error, and I would not use the err to direct the flow here. It is basically the same reason why a map has `value, ok := foo[asd]` and not `value, err := foo[asd]`

Teelevision: There is probably no reason for or against either version, except taste. I like it with the two return values.

threez: I like the idea, will add.

http/longpoll/longpoll.go
threez: So you argue to remove the c.RetryTime to stay in the time budget, like `if !time.Now()Add(c.RetryTime).Before(until) { `

http/longpoll/longpoll.go
Teelevision: I second the idea of passing the context to the `LongPollFunc`.

Smuerdt: But with this approach the caller needs to check the bool and the err and can't just check the error if he does not event want to retry it. It's like omitting the bool on the map example. (If you don't care about existence and can also work with the default)

http/longpoll/longpoll.go
Teelevision: That should work as well.

http/longpoll/longpoll.go
threez: I don't like to handle an error in this case but I agree, that the caller might not understand the behavior. It needs to be restricted to avoid long hanging go-routines in http handlers and the like. So instead of dropping the arg silently the value will be set to min/max if duration is given and otherwise out of bounds

http/longpoll/longpoll.go
Teelevision: But the caller is this code and this code wants to check both the error and retry case.

http/longpoll/longpoll.go
Teelevision: Please also update the comment of both functions that there are boundaries and that this is the behaviour.

http/longpoll/longpoll.go
Smuerdt: I was talking about the whole approach and it seemed unnecessary to me to open the same discussion on `Until(...) (ok bool, err error)` and `LongPollUntil(...) (ok bool, err error)`...

Teelevision: Oh ok, in that case:

> It's like omitting the bool on the map example.

```ok, _ := longpull.Until(...)```

Comment (Smuerdt) [reply http/longpoll/longpoll.go:11]: More like `_, err := longpull.Until(...)` which brings me to my point. With a map you can do `val := m[0]` as it is a builtin feature to omit the ok, which is not possible for functions and as I don't see a valid usecase for `true, error` or `false, nil` I already see that we always omit the bool. However the other comments are more important and we should not waste to much time on this discussion. If you want to have it as is, keep it.

File: http/longpoll/longpoll.go (hunk 1 above)
Comment (threez) [reply http/longpoll/longpoll.go]: Yes, did so.

File: http/longpoll/longpoll.go (hunk 2 above)
Comment (threez) [reply http/longpoll/longpoll.go]: Please check the current MR.

File: http/longpoll/longpoll.go [hunk 4]
```diff
@@ -0,0 +1,92 @@
...
+
+		// long pooling desired?
+		if !time.Now().Add(c.RetryTime).Before(until) {
+			break
+		}
+
+		select {
+		// handle context cancelation
+		case <-ctx.Done():
+			return false, ctx.Err()
```
Teelevision, line 85: This would now return a context error, but it's not necessarily related to the passed in context. Which is only confusing to the caller, because the caller might get a context error even though the passed in context is valid.

PR #209, opened 2020-06-22
Add queueing system via rmq

Ferlonas:


backend/queue/rmq.go
```diff
@@ -0,0 +1,64 @@
...
+	"time"
+
+	"github.com/adjust/rmq/v2"
+	"github.com/pace/bricks/backend/redis"
+	"github.com/pace/bricks/maintenance/health/servicehealthcheck"
+)
+
+var (
+	rmqConnection     rmq.Connection
+	queueHealthLimits map[string]int
```
threez, line 15: Since we have two different go routines that can read/write we should use `sync.Map`

Ferlonas: Done

PR #210, opened 2020-06-25
Prevent panic on queue package loading

Ferlonas:


backend/queue/rmq.go
```diff
@@ -16,7 +16,10 @@ var (
 	queueHealthLimits sync.Map
 )
 
-func init() {
+func initDefault() {
+	if rmqConnection != nil {
```
daemonfire300, line 20: Singleton pattern?

daemonfire300: Let me rephrase that: Am I seeing a singleton pattern here, and if yes is this intentional, and if yes why?

Ferlonas: As only one connection to redis is required for the queue to work, I intended to do this on package loading. Unfortunately, this crashes the unit tests, as the init function is called even if the package is not used, only imported.

I'm switching to this way of doing it as I don't want to have the caller call the initialization first, and I do not want to wrap it in another struct, which I would then create with NewConnection(), whose only purpose would be to initialize this stuff.

daemonfire300: Is there a functionality to "close" the queue and does that also close the connection to redis?

Ferlonas: There are ways to close the queue, but as a connection creates multiple queues, this will not close the connection itself.

daemonfire300: Ok, so I am accepting this, but I do not have the best of feelings because I suspect we might run into problems reusing the same redis connection for every queue if this should ever be used with multiple queues and multiple routines (multiple redis servers).

PR #217, opened 2020-08-12
add cache package

Teelevision:
I extracted the cache from one of our services, because I thought this was already a pretty general solution that we could make use of in other places as well.

I'm a bit unsure about the naming. Because the package is already called `cache` I tried to avoid using cache, like `cache.Cache` and `cache.RedisCache`. So a struct that needs a cache could look like this:
```go
type NameMap struct {
    Names cache.Strings
}
```
And initialisation like this:
```go
nm.Names = cache.InMemory()
// or
nm.Names = cache.InRedis(redisClient, "prefix")
```


pkg/cache/cache.go
```diff
@@ -0,0 +1,26 @@
...
+	"time"
+)
+
+// Strings caches strings.
+type Strings interface {
+	// SetString stores the value under the key. Any existing value is
+	// overwritten. If forgetAfter is given, the cache automatically forgets the
+	// value after the duration. If forgetAfter is zero then it is never
+	// automatically forgotten.
+	SetString(ctx context.Context, key, value string, forgetAfter time.Duration) error
```
daemonfire300, line 17: why not use commonly used terminology like `TTL` https://en.wikipedia.org/wiki/Time_to_live ?

pkg/cache/cache.go
```diff
@@ -0,0 +1,26 @@
...
+
+	// GetString returns the value stored under the key. If there is no value
+	// stored, ErrNotFound is returned.
+	GetString(ctx context.Context, key string) (string, error)
```
daemonfire300, line 21: Wouldn't it be nice to also return how long this item will be valid, i.e., return the ttl/forgetAfter. Because if I fetch an item from the cache that becomes invalid in 5 minutes but I expect my process to run 10 minutes I might want to check for how long the value can be considered valid.

Otherwise I would need to re-fetch the item continuously to check whether it changed or is invalid/gone

pkg/cache/cache.go
Teelevision: good idea

pkg/cache/cache.go
Teelevision: It would be possible in both the memory and [redis](https://stackoverflow.com/questions/46064445/equivalent-of-setex-for-get-and-ttl) implementation. I'll think about it.

pkg/cache/cache.go
```diff
@@ -0,0 +1,26 @@
...
+
+package cache
+
+import (
+	"context"
+	"time"
+)
+
+// Strings caches strings.
+type Strings interface {
```
threez, line 12: I would change the interface to operate on bytes, like most other kv stores do:

* https://godoc.org/github.com/cznic/kv
* https://pkg.go.dev/bazil.org/bazil/kv?tab=doc
* https://pkg.go.dev/github.com/syndtr/goleveldb/leveldb?tab=doc
Concrete: 

```go
Get(ctx context.Context, key []byte) ([]byte, error)
Put(ctx context.Context, key, value []byte) error
PutTTL(ctx context.Context, key, value []byte, ttl time.Duration) error
```

pkg/cache/cache.go
threez: That should be a separate func if it involves calling redis twice

Teelevision: I implemented the change.

pkg/cache/cache.go
Teelevision: Done.

pkg/cache/cache.go
Teelevision: We are free to add another interface, that works on bytes, if needed. But this one works on strings. I see this package as a convenience layer on top of other key-value stores. There wouldn't be much convenience if I still need to cast everything.

Teelevision:
I added a test suite.

pkg/cache/cache.go
Teelevision: I added some package documentation about the purpose of this package.

pkg/cache/redis_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/pace/bricks/pkg/cache/testsuite"
+	"github.com/stretchr/testify/suite"
+)
+
+func TestIntegrationRedis(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+	suite.Run(t, &testsuite.StringsTestSuite{
+		Cache: cache.InRedis(redis.Client(), "test:cache:"),
```
Smuerdt, line 20: I think you should flush all redis keys with the given prefix before and after the test to make the test more robust against multiple re runs and parallel execution.

Teelevision: How do I do that?

Teelevision: I found this: https://stackoverflow.com/questions/4006324/how-to-atomically-delete-keys-matching-a-pattern-using-redis

pkg/cache/cache.go
Teelevision: Ok, the following problem exists with my idea. If for example the Redis implementation should support both strings and ints, then it anyway needs some form of abstraction to distinguish strings from strings that encode ints. And one would be able to request a string as an int, which could fail. But this error category would be unclear to the caller. Also one could store a string containing a number and request it as int. So it works even though it shouldn't. I don't like this error category, so I will think about the bytes approach once more.

pkg/cache/testsuite/strings.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+}
+
+func (suite *StringsTestSuite) TestGetString() {
+	c := suite.Cache
+	ctx := log.WithContext(context.Background())
+
+	_ = c.Forget(ctx, "foo") // make sure it doesn't exist
+	suite.Run("returns the ttl if set", func() {
+		_ = c.SetString(ctx, "foo", "bar", time.Minute)
```
Teelevision, line 73: I should use different keys in every test that can run synchronously.

pkg/cache/redis_test.go
Teelevision: I clean up every key in the tests, so I think this should be fine.

pkg/cache/cache.go
Teelevision: I'll change the interface to use `string` for the key and `[]byte` for the value.

pkg/cache/testsuite/strings.go
Teelevision: No need, testify runs everything in sequence unless requested otherwise. See https://github.com/stretchr/testify/issues/187.

pkg/cache/cache.go
Teelevision: Done.

PR #230, opened 2020-09-25
Sort stable everywhere

Smuerdt:


http/jsonapi/runtime/error.go
```diff
@@ -118,6 +116,7 @@ func WriteError(w http.ResponseWriter, code int, err error) {
 		if source := ei.Source; source != nil {
 			ev = ev.Fields(*source)
 		}
+		ev.Msg("")
```
Smuerdt, line 119: Had a linter issue that ev is unused. Is this what you intended here @threez

threez: Hmm, yes the event is build based on the parameters and then issued with `.Msg(...)`. Why is the linter complaining?

Smuerdt: I added the `ev.Msg`. This code was not there before that's why the linter complained that `ev` was unused.

threez: We should delete the loop to avoid multiple info lines or only log on debug level. The msg should not be empty, but contain at least `.Msg("Send error to client")`

Smuerdt: done

http/jsonapi/runtime/error.go
```diff
@@ -79,46 +78,10 @@ func (e Errors) setID(errorID string) {
...
-	errList.List.setID(reqID)
-
-	// render the error to the client
-	enc := json.NewEncoder(w)
-	enc.SetIndent("", "  ")
-	err = enc.Encode(errList)
-	if err != nil {
-		log.Logger().Info().Str("req_id", reqID).
-			Err(err).Msg("Unable to send error response to the client")
-	}
```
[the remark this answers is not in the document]
Smuerdt: Please move discussion to https://github.com/pace/bricks/pull/231

http/jsonapi/runtime/error.go
Smuerdt: Please move discussion to #231

Issue #235, opened 2020-10-21
objstore: move healthcheck registration into client creation


daemonfire300:
# Motivation

Do not register healthchecks if the package is simply being imported but the client not necessarily used, i.e., move them out of the `init()` method into the client creation.

PR #236, opened 2020-10-21
Objstore: Remove healthchecks from init() and wrap in sync.Once #235

daemonfire300:
please review and merge

Closes https://github.com/pace/bricks/issues/235

### DIFF
--- a/backend/objstore/objstore.go
+++ b/backend/objstore/objstore.go
@@ -2,11 +2,13 @@ package objstore
 
 import (
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/caarlos0/env"
 	"github.com/minio/minio-go/v6"
 	"github.com/minio/minio-go/v6/pkg/credentials"
+
 	"github.com/pace/bricks/http/transport"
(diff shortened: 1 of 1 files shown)


backend/objstore/objstore.go
```diff
@@ -63,6 +43,12 @@ func Client() (*minio.Client, error) {
 	})
 }
 
+// Client with environment based configuration. Registers healthchecks automatically. If yo do not want to use healthchecks
+// consider calling CustomClient.
+func DefaultClientFromEnv() (*minio.Client, error) {
+	return Client()
```
Ferlonas, line 49: As you did create a client() function, I think it would be prudent to use that one here instead of the old soon-to-be-deprecated Client()

Ferlonas (changes requested):

backend/objstore/objstore.go
daemonfire300: @Ferlonas addressed. please re-review

Ferlonas (approved):

PR #237, opened 2020-10-26
Make postgres config public to increase re-useability

daemonfire300:
Motivation:

By exporting the Config struct, packages that want to construct an additional or custom pg pool
can now rely on default set via ENVs that are/should be set anyway.
Those implementations can then pick selectivly which defaults they want and which to override.


roberth1988 (approved):

daemonfire300:
@Smuerdt ping

Smuerdt (approved):

PR #243, opened 2020-12-16
add redaction system package #238

### LINKED ISSUE #238: Extend http DumpRoundTripper to allow hiding of sensitive headers
# Motivation

provide sane settings for when the roundtripper is used in production if needed.

# Proposed Solution

Add `noauth` config option and or header white/blacklist to filter out sensitive headers.

threez:
closes #238


threez:
Test only fail because of docker pull limit.

http/transport/dump_round_tripper.go
```diff
@@ -114,12 +132,19 @@ func (l *DumpRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
...
+
 	// response logging
 	if l.DumpResponse || l.DumpResponseHEX {
 		respDump, err := httputil.DumpResponse(resp, l.DumpBody)
 		if err != nil {
 			respDump = []byte(err.Error())
 		}
+
+		// in case a redactor is present, redact the content before logging
+		if redactor != nil {
```
daemonfire300, line 144: Minor: We could save a function call by checking whether `len(respDump) > 0` or something, to avoid calling it on an empty string

pkg/redact/redact.go
```diff
@@ -0,0 +1,62 @@
...
+// NewPatternRedactor creates a new redactor for masking certain patterns
+func NewPatternRedactor(scheme RedactionScheme) *PatternRedactor {
+	return &PatternRedactor{
+		scheme: scheme,
+	}
+}
+
+func (r *PatternRedactor) Mask(data string) string {
+	for _, pattern := range r.patterns {
+		if pattern == nil {
```
daemonfire300, line 24: Why do we even allow for pattern to be nil, is there a purpose for this?

daemonfire300 (commented):
LGTM but two comments/questions.

http/transport/dump_round_tripper.go
threez: the respDump is never nill (it is either the complete response or the error message). IMHO the check would not be working because of that.

pkg/redact/redact.go
threez: Technically one could add a nil pattern using the `AddPatterns`.

daemonfire300: Resolving this issue, because thinking about it again, it's a technicality and should not pose a real word problem

daemonfire300 (approved):

PR #246, opened 2020-12-30
Log failed health checks

Ferlonas:
Also report number of errors and warnings in failure case

### DIFF
--- a/maintenance/health/servicehealthcheck/health_handler.go
+++ b/maintenance/health/servicehealthcheck/health_handler.go
@@ -4,6 +4,7 @@
 package servicehealthcheck
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/pace/bricks/maintenance/log"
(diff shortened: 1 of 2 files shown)


pnull:
Not sure if I understand the code correctly, but just to make sure: The log should help us with identifying the root cause of failed health checks. Therefore, it should not just log "something failed" but details about which check exactly failed and why. Example: "database connection timeout" or "could not resolve database host name" or "database write operation failed: %s". Something like that ;)

Ferlonas:
@pnull Unfortunately at this part of the code we don't have the knowledge what kind of check it executed. However, we do have the error or warning message, all of which are stored and then logged with the `Strs("errors", errors)` call. For example, a log message may look like this:
```
{"level":"info","warnings":[],"errors":["Postgres connection error: no route to host","could not resolve host 'redis'],"time":"2020-12-30T12:39:44Z","message":"Health check failed"}
```

pnull:
@Ferlonas awesome. As expected: I didn't understand the code correctly. Your solution is perfectly fine! ;-)

PR #249, opened 2021-01-08
248 s3 health versions

threez:


backend/objstore/health_objstore.go
```diff
@@ -52,18 +53,38 @@ func (h *HealthCheck) HealthCheck(ctx context.Context) servicehealthcheck.Health
...
+		},
 	)
 	if err != nil {
 		h.state.SetErrorState(fmt.Errorf("failed to get object: %v", err))
 		return h.state.GetState()
 	}
 	defer obj.Close()
 
+	// in case the bucket is versioned, delete versions to not leak versions
+	if info.VersionID != "" {
```
daemonfire300, line 73: This only downloads the last version, correct?

threez: It only deletes objects when they are versioned.

daemonfire300: Yes but only the `latest` version, because the above code would only return the `latest` version, correct? That's what my question was about.

daemonfire300 (changes requested):
LGTM

daemonfire300 (approved):
LGTM

PR #251, opened 2021-01-18
Add a configurable maximum size for log sinks

Ferlonas:


threez (changes requested):

maintenance/log/sink.go
```diff
@@ -45,13 +47,27 @@ func SinkContextTransfer(sourceCtx, targetCtx context.Context) context.Context {
 // logs, created with log.Ctx(ctx), inside the context
 // and use them at a later point in time
 type Sink struct {
-	Silent       bool
-	jsonLogLines []string
+	Silent     bool
+	CustomSize int
+	ring       *stringRing
```
threez, line 52: I think we should allocate the ring on the stack and avoid the pointer, that should then also prevent the init

maintenance/log/sink.go
```diff
@@ -126,3 +153,70 @@ func (s *Sink) Write(b []byte) (int, error) {
 
 	return s.output.Write(b)
 }
+
+func (s *Sink) initialize() {
```
threez, line 157: I think we should not initialize the buffer with a fixed size before, this will be very costly for every HTTP request. I prefer the logic from before where the append is growing the buffer on demand. But this buffer should be kept at either defaultSinkSize or CustomSize. So I would prefer to keep a max value.

threez (changes requested):

maintenance/log/sink.go
Ferlonas: It's not especially costly. Each uninitialized string has a memory use of 2*sizeof(int). That's something I believe we can afford. I prefer this to the additional time added by an append() call and the resulting complication in the implementation.

maintenance/log/sink.go
Ferlonas: I can remove the pointer, although that may not necessarily allocate it from the stack. The init on the other hand is for ensuring thread safety, and for enabling us to improve on the functionality in the future without having to work around it, like we do now.

A sink should be initialized with a certain size which is not allowed to change after creation.

daemonfire300:
The following assumes that we are using a fixed-size and finite-size data structure.

Would it make sense to reserve "the last" line in the whatever (fixed) sized (ring)buffer for the following case:

We hit `n-1` messages, naively we would start to overwrite (a) at the beginning of the ring (b) truncate the buffer and write to the newly freed space.

What we would not see in the logs if that happened, unless we see an abrupt cut. Would it therefore make sense to log an indicator line at the end, e.g., `warning: buffer reached maximum size`.

Not sure though if a ring is the right choice. I thought it would be, but now I am not that sure anymore, because we do not want to re-use memory, i.e., we do not want to lose log information.

PR #252, opened 2021-01-26
Use dump round tripper as last in the chain

Ferlonas:


daemonfire300 (approved):
Did we test this change? Looks good, but we should ensure that it has no side effects.

daemonfire300:
@Ferlonas ping

PR #254, opened 2021-03-02
log errors that are send to the http clients

threez:


daemonfire300:
My only concern is the commit message, at first I was confused what it meant.

Maybe `log errors that are send from the server to the http clients` or `log errors when writing response to http clients (api,runtime)`

daemonfire300 (approved):
See comment on MR

http/jsonapi/runtime/error.go
```diff
@@ -98,9 +98,14 @@ func WriteError(w http.ResponseWriter, code int, err error) {
...
 
+	reqID := w.Header().Get("Request-Id")
+
 	// update the http status code of the error
 	errList.List.setHTTPStatus(code)
-	reqID := w.Header().Get("Request-Id")
+
+	// log all errors send to clients
+	log.Logger().Debug().Str("req_id", reqID).
+		Err(errList.List).Msg("error send to client")
```
Ferlonas, line 108:
```suggestion
		Err(errList.List).Msg("error sent to client")
```

PR #276, opened 2021-09-03
add grpc client and server middleware setup code

threez:
Adds a new grpc client and server middleware setup that works together with the rest of pace bricks.

* adds testserver for testing and demonstration
* fixes golint to use vendored version (prevents crashes)
* updates readme


daemonfire300:
I added DialContext.

Ferlonas (approved):

PR #277, opened 2021-09-30
Improve jaeger tracing

roberth1988:
- Removed the full SQL Query from the operation name
- Add tag to be able to filter things easier

### DIFF
--- a/backend/postgres/postgres.go
+++ b/backend/postgres/postgres.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


backend/postgres/postgres.go
```diff
@@ -271,10 +271,11 @@ func openTracingAdapter(event *pg.QueryProcessedEvent) {
 		q = qe.Error()
 	}
 
-	name := fmt.Sprintf("PostgreSQL: %s", q)
-	span, _ := opentracing.StartSpanFromContext(event.DB.Context(), name,
+	span, _ := opentracing.StartSpanFromContext(event.DB.Context(), "sql:query",
```
threez, line 274: I would add the first term e.g. INSERT, SELECT, ... to he query as well. e.g. `sql:insert`

backend/redis/redis.go
```diff
@@ -170,10 +169,12 @@ type logtracerValues struct {
 func (lt *logtracer) BeforeProcess(ctx context.Context, cmd redis.Cmder) (context.Context, error) {
 	startedAt := time.Now()
 
-	span, _ := opentracing.StartSpanFromContext(ctx, fmt.Sprintf("Redis: %s", cmd.Name()))
+	span, _ := opentracing.StartSpanFromContext(ctx, cmd.Name())
```
threez, line 172: should this then be `"redis:" + cmd.Name()` ?

threez (changes requested):
Please also fix the tests, they broke.

PR #283, opened 2021-11-09
couchdb: Switch to less spammy couchdb default configuration by NOT i…

daemonfire300:
…nheriting configuration from ENV by default

First step to selectively reduce log-spam when fetching data.

### DIFF
--- a/http/transport/dump_round_tripper.go
+++ b/http/transport/dump_round_tripper.go
@@ -5,8 +5,11 @@ package transport
 
 import (
 	"encoding/hex"
+	"fmt"
 	"net/http"
 	"net/http/httputil"
+	"net/url"
+	"strings"
 
 	"github.com/caarlos0/env"
 	"github.com/pace/bricks/maintenance/log"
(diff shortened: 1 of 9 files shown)


daemonfire300:
Solution is not the greatest. Ideally I would like to somehow make this generic and more flexible without specific configuration.

threez:
I see the problme and agree that we want a more generic solution. I would prefer if we can adjust the configuration for http dumps, so lets think about it this way:

I enabled full logging for a service:

```
HTTP_TRANSPORT_DUMP=request,response,request-hex,response-hex,body
```

But i want to overwrite the configuration for couch or any other http client that has large bodies, we could write:

```
HTTP_TRANSPORT_DUMP_DISABLE_URL_PREFIX=request-hex,response-hex,body http://couch.local http://big.body.com
```

Idea being that we can reduce parts of the configuration (e.g. body printing for certain url prefixes without affecting other logs while having only one implementation in the transport.

daemonfire300:
> I see the problme and agree that we want a more generic solution. I would prefer if we can adjust the configuration for http dumps, so lets think about it this way:
> 
> I enabled full logging for a service:
> 
> ```
> HTTP_TRANSPORT_DUMP=request,response,request-hex,response-hex,body
> ```
> 
> But I want to overwrite the configuration for couch or any other http client that has large bodies, we could write:
> 
> ```
> HTTP_TRANSPORT_DUMP_DISABLE_URL_PREFIX=request-hex,response-hex,body http://couch.local http://big.body.com
> ```
> 
> Idea being that we can reduce parts of the configuration (e.g. body printing for certain url prefixes) without affecting other logs while having only one implementation in the transport. In this example only `HTTP_TRANSPORT_DUMP=request,response` would be effective for `http://couch.local` and `http://big.body.com`

I see your point, but when do you realistically need to do that (limiting to urls).
This actually puts more burden on the developer to configure a dump body blacklist, and given that URLs vary from environment to environment this would require an extra configuration step per environment. It kinda does already, though you now introduce the possibility to make a typo in the config in each new environment. And if the URLs change you are also required to change the configuration, although for you as a developer nothing changed: You still do not want to dump the body.

Is the motivation of being extra-selective differentiating between seemingly small `PUT` calls, so that one can inspect what is being written to CouchDB while we want to ignore `GET` calls? E.g., disable on `/dev-some-db/_find`?

daemonfire300:
Will change

daemonfire300:
Going to push adjusted version that mostly aligns with the proposal tomorrow (hopefully)

Ferlonas (approved):

PR #285, opened 2021-11-26
Make errors.Hide no longer prepend the external error message

Ferlonas:
This was deemed too confusing as the prepended error message is expected to define the
error type, but in this case it keeps being a context error


srieckhof:
Closing this PR because it's old and we are missing the context.

PR #290, opened 2021-12-23
remove log entry to reduce redis log entries

pieceofsoul:


daemonfire300:
why are we removing this?

roberth1988:
log spam, or we should differentiate in log levels more.

daemonfire300:
This log line is already labelled as debug and we still want this information, dont we?

roberth1988:
Then we should rethink how we run some services in our internal infra ;)

pieceofsoul:
There is a metric covering the same information as the log line, beside the potential error. Maybe adding the log line in the error case would be an improvement and cover @daemonfire300 concerns.

PR #295, opened 2022-01-24
Readiness failover

threez:


pieceofsoul:
On the official redis page there is a different DLM proposed for go. Should we switch to that?

https://redis.io/topics/distlock

https://github.com/go-redsync/redsync

maintenance/failover/failover.go
```diff
@@ -0,0 +1,245 @@
...
+			return ctx.Err()
+		case <-a.close:
+			return nil
+		case <-t.C:
+			if a.getState() == ACTIVE {
+				err := lock.Refresh(a.timeToFailover, &redislock.Options{
+					RetryStrategy: redislock.LimitRetry(redislock.LinearBackoff(a.timeToFailover/3), 3),
+				})
+				if err != nil {
+					logger.Debug().Err(err).Msgf("failed to refresh")
```
pieceofsoul, line 134:
```suggestion
					logger.Debug().Err(err).Msg("failed to refresh")
```

maintenance/failover/failover.go
```diff
@@ -0,0 +1,245 @@
...
+				}
+
+				// lock acquired
+				logger.Debug().Msg("becoming active")
+				a.becomeActive(ctx)
+
+				// we are active, renew if required
+				d, err := lock.TTL()
+				if err != nil {
+					logger.Debug().Err(err).Msgf("failed to get TTL")
```
pieceofsoul, line 162:
```suggestion
					logger.Debug().Err(err).Msg("failed to get TTL")
```

maintenance/failover/failover.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (a *ActivePassive) becomeUndefined(ctx context.Context) {
+	a.setState(ctx, UNDEFINED)
+}
+
+// setState returns true if the state was set successfully
+func (a *ActivePassive) setState(ctx context.Context, state status) bool {
+	err := a.client.SetCurrentPodLabel(ctx, Label, a.label(state))
+	if err != nil {
+		log.Ctx(ctx).Error().Err(err).Msgf("failed to mark pod as undefined")
```
pieceofsoul, line 228:
```suggestion
		log.Ctx(ctx).Error().Err(err).Msg("failed to mark pod as undefined")
```

threez:
> On the official redis page there is a different DLM proposed for go. Should we switch to that?
> 
> https://redis.io/topics/distlock
> 
> https://github.com/go-redsync/redsync

In case we want to switch the lock implementation, we should do it in a sep. issue as this would also affect the routine pkg.

PR #296, opened 2022-01-28
new round tripper package

pieceofsoul:
The old roundtripper package did not preserve the request body on retries. This has been solve by exchanging to another package providing that handling.


daemonfire300:
Lgtm

http/transport/retry_round_tripper.go
```diff
@@ -77,17 +76,11 @@ func (l *RetryRoundTripper) SetTransport(rt http.RoundTripper) {
...
 
-	var maxRetriesError retry.MaxError
 	if err != nil {
-		switch {
-		case errors.As(err, &maxRetriesError):
-			return nil, ErrRetryFailed
-		default:
-			return nil, err
-		}
+		return nil, err
```
threez, line 83: We should keep mapping the `ErrRetryFailed` as code may depends on it.

http/transport/retry_round_tripper_test.go
```diff
@@ -7,107 +7,129 @@ import (
...
-	body string
 	// returned error
 	err error
 	// recorded context
 	ctx context.Context
 }
 
 func (t *retriedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	t.attempts++
+	fmt.Printf("Request Nr %d %d\n", t.attempts, t.statusCodes[t.attempts])
```
threez, line 123: pls remove the print

threez (changes requested):

http/transport/retry_round_tripper.go
pieceofsoul: The used package does not provide a dedicated error on exceeded retry. Instead the normal request error / status is returned. Currently I do not see an easy way to achieve this.

http/transport/retry_round_tripper_test.go
pieceofsoul: Has been removed.

daemonfire300:
Pipeline red?

pieceofsoul:
Closing in favour of https://github.com/pace/bricks/pull/326

PR #299, opened 2022-02-16
Use log.Fatal when JAEGER_SERVICE_NAME is not present

arnold-iakab:


backend/postgres/postgres.go
```diff
@@ -137,7 +137,6 @@ func init() {
 	if cfg.ApplicationName == "-" {
 		names := []string{
 			os.Getenv("JAEGER_SERVICE_NAME"),
-			os.Getenv("SERVICE_NAME"),
```
daemonfire300, line 140: This is safe to remove, because we never actually used it?

maintenance/tracing/tracing.go
```diff
@@ -31,8 +31,7 @@ func init() {
 		return
 	}
 	if cfg.ServiceName == "" {
-		log.Warn("Using Jaeger noop tracer since no JAEGER_SERVICE_NAME is present")
-		return
+		log.Fatal("JAEGER_SERVICE_NAME is not present")
```
daemonfire300, line 34: The advantage is: we get a non-zero exit code of the program now?

pieceofsoul: I would not change the behaviour to fatal since this is called in an init function what can have unexpected behaviour changes at other places. If a missing `ServiceName ` results in a program failure, we should instead make the original warning a valid case ("Using noop ...")

backend/postgres/postgres.go
pieceofsoul: This overall logic is highly confusing. I would rather keep the `SERVICE_NAME ` and in case, none is given log a warning that the `JAEGER_SERVICE_NAME` was used as fallback option to make the behaviour more transparent to the user.

maintenance/tracing/tracing.go
daemonfire300: Then let's restructure it completely and dump the code in init() and move it somewhere explicit?

PR #305, opened 2022-04-21
Fix queue backend metrics that were gathered only once

arnold-iakab:
Metrics gathered by the queue backend were not working as expected due to the fact that they were gathered only once on package initialization. Added a ticker to the routine that is responsible with gathering the metrics.


daemonfire300:
nananan tests failed? :D

PR #306, opened 2022-04-22
log proper failover state on error

mihailozarinschi:
During active/passive failover, if setting the label of the pod fails, the logged message always contains the state `undefined` and not the actual state it tried to set.


daemonfire300:
Running CI and then merging. LGTM

daemonfire300:
hm retriggered still failing 🤔

PR #307, opened 2022-04-26
Run healthchecks in the background

mihailozarinschi:
Run HealthCheck in the background with configurable interval and timeout (max wait).


maintenance/health/servicehealthcheck/background_healthchecker.go
```diff
@@ -0,0 +1,89 @@
...
+		if initHC, ok := bhc.(Initializable); ok {
+			if initErr = initHC.Init(ctx); initErr != nil {
+				log.Warnf("error initializing background health check %q: %s", name, initErr)
+				bgState.SetErrorState(initErr)
+			}
+		}
+		timer := time.NewTimer(0) // Do first health check instantly
+		for {
+			func() {
+				<-timer.C
```
daemonfire300, line 43: can we move this out of the function, because we do not need to start a func when it just waits. Does not really matter in the grand scheme, but it's a function sitting on the call stack for N minutes when it could only sit there when it is actually needed.

maintenance/health/servicehealthcheck/background_healthchecker.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// registerBackgroundHealthCheck will run the HealthCheck in the background at every given interval.
+// The returned backgroundStateHealthChecker is then used to query this cached state when needed.
+func registerBackgroundHealthCheck(name string, bhc BackgroundHealthCheck) *backgroundStateHealthChecker {
+	var bgState ConnectionState
+
+	go func() {
+		s := log.Sink{Silent: true}
+		ctx := log.ContextWithSink(context.Background(), &s)
```
daemonfire300, line 31: Have you tried running this code without any initialization of any logger. Because I think if this runs isolated from any other contexts if no logger is attached to the context then no logging actually happens. That is at least my observation when I forget to attach anything to the context regarding logging.

But I am blissfully unaware of the internals of our zerolog wrapper.

maintenance/health/servicehealthcheck/background_healthchecker.go
```diff
@@ -0,0 +1,89 @@
...
+type BackgroundHealthCheckMaxWait interface {
+	MaxWait() time.Duration
+}
+
+// registerBackgroundHealthCheck will run the HealthCheck in the background at every given interval.
+// The returned backgroundStateHealthChecker is then used to query this cached state when needed.
+func registerBackgroundHealthCheck(name string, bhc BackgroundHealthCheck) *backgroundStateHealthChecker {
+	var bgState ConnectionState
+
+	go func() {
```
daemonfire300, line 29: As far as I can see this routine does not have any panic recovery, right? Are we fine with it dying and not reporting any results anymore and then see that because no results are being reported?

maintenance/health/servicehealthcheck/background_healthchecker.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+		}
+		timer := time.NewTimer(0) // Do first health check instantly
+		for {
+			func() {
+				<-timer.C
+				defer timer.Reset(bhc.Interval())
+
+				maxWait := cfg.HealthCheckMaxWait
+				if mw, ok := bhc.(BackgroundHealthCheckMaxWait); ok {
```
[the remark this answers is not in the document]
mihailozarinschi: The current implementation of this packages is based on the the following interface:
```go
type HealthCheck interface {
	HealthCheck(ctx context.Context) HealthCheckResult
}
```

At the moment, to add **optional**, but also backward compatible, behavior, the implementation has to implement additional interfaces, eg: 
```go
type Initializable interface {
	Init(ctx context.Context) error
}
```
which allows to have custom initialisation if needed by a specific HealthCheck.

The `BackgroundHealthCheck` follows the same principle, and it contains only the method `Interval() time.Duration` because this is the only one required to make it run in the background. 

The maxWait is optional because the package itself already has a generic maxWait config for all HealthChecks, and to not force every background HC to provide a maxWait method, it was done via another interface that can be implemented as/when needed.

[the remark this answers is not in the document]
mihailozarinschi: Another point to why this might be a better solution than a config variable passed at the beginning, is that having a method, that is called each time, allows the implementation to provide dynamic `Interval` but also `MaxWait` depending on it's current state.

maintenance/health/servicehealthcheck/background_healthchecker.go
mihailozarinschi: Yep, no need for it here as it uses the `context.Background()`.  

I copied this from the other function that had this, but that was called from an `http.Handler` with an active logger in the context so it made sense to mute it.

daemonfire300:
Will revisit this on monday 👍

daemonfire300:
@Ferlonas since this is related to an internal issue and supposed to improve a certain scenario can you also take look at it?

maintenance/health/servicehealthcheck/background_healthchecker.go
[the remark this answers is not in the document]
daemonfire300: Ok.

daemonfire300:
So for SOME reason this PR does not have a problem with the CI but Arnolds PR does not work :/ dang github.dockerhub ci

maintenance/health/servicehealthcheck/background_healthchecker.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/opentracing/opentracing-go"
+
+	"github.com/pace/bricks/maintenance/errors"
+	"github.com/pace/bricks/maintenance/log"
+)
+
+// BackgroundHealthCheck is an interface that, when implemented, will run the
+// HealthCheck in the background, and cache the result, that will be fetched by the HealthHandler
+type BackgroundHealthCheck interface {
```
Ferlonas, line 16: I assume you did this in the assumption that we want to give each health check the option of defining its own intervals. However, that would require us to change every health check implementation to support the Interval() function.

Instead, I propose to add Options to the `RegisterHealthCheck` and `RegisterOptionalHealthCheck` functions that give the owner the option of setting an interval, and using a struct to store it here.

Ferlonas (changes requested):

maintenance/health/servicehealthcheck/background_healthchecker.go
mihailozarinschi: Added the Options. 

Although, one thing I'm not sure of, is the use of the MaxWait in the HealthChecks called from the Handler. 
Basically the `cfg.HealthCheckMaxWait` is the default, and each HealthCheck can have its own, which is fine for background ones, but might not be fine for those called from the Handler, as the bigger maxWait becomes a maxWait for the entire Handler.

Although, if not using it this way, and having the `cfg.HealthCheckMaxWait` as a strong timeout on the Handler, then `UseMaxWait` is an option usable for background HealthChecks only, which could look a bit strange from the package API point of view.

Ferlonas (approved):

PR #310, opened 2022-06-07
Extract limited stack trace frames from error

arnold-iakab:


maintenance/terminationlog/termlog.go
```diff
@@ -9,16 +9,27 @@ package terminationlog
...
 import (
 	"fmt"
+	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 	"os"
 )
 
 var logFile *os.File
 
+const StackTraceLimit = 100
```
daemonfire300, line 19: why exactly 100?

PR #311, opened 2022-06-08
Limit termination log output

arnold-iakab:


daemonfire300:
@arnold-iakab do we still take this with us?

PR #313, opened 2022-06-27
pg: make query logging configurable

daemonfire300:


backend/postgres/postgres.go
```diff
@@ -228,15 +232,50 @@ func CustomConnectionPool(opts *pg.Options) *pg.DB {
...
+
+const (
+	readMode  queryMode = iota
+	writeMode queryMode = iota
+)
+
+// determineQueryMode is a poorman's attempt at checking whether the query is a read or write to the database.
+// Feel free to improve.
+func determineQueryMode(qry string) queryMode {
+	if strings.HasPrefix(strings.ToLower(qry), "select") {
```
daemonfire300, line 257: could probably benefit from a LeftTrim too

backend/postgres/postgres.go
```diff
@@ -228,15 +232,50 @@ func CustomConnectionPool(opts *pg.Options) *pg.DB {
 		Str("as", opts.ApplicationName).
 		Msg("PostgreSQL connection pool created")
 	db := pg.Connect(opts)
-	db.OnQueryProcessed(queryLogger)
+	if cfg.LogWrite || cfg.LogRead {
+		db.OnQueryProcessed(queryLogger)
+	} else {
+		log.Logger().Warn().Msg("Connection pool has logging queries disabled completely")
```
jpwenzel, line 238: Is this an information that we would want to log on `Info` level, assuming it happens only once per opened connection?

backend/postgres/postgres.go
jpwenzel: Use `TrimSpace` maybe?

daemonfire300: Done

pieceofsoul (approved):

PR #324, opened 2022-09-23
Configure log handler using env variable

arnold-iakab:


maintenance/log/handler.go
```diff
@@ -28,10 +28,18 @@ const RequestIDHeader = "Request-Id"
 // in the request specific Sink.
 func Handler(silentPrefixes ...string) func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
-		return hlog.NewHandler(log.Logger)(
+		chain := hlog.NewHandler(log.Logger)(
```
daemonfire300, line 31: I think we can skip the variable assign and just return this if TRUE und else the part that is in the if guard you added, right?

daemonfire300 (changes requested):

maintenance/log/handler.go
arnold-iakab: Fixed

PR #329, opened 2022-11-08
remove pace gitlab urls, upgrade dependencies

flofuenf:
Due to a change in the infrastructure, the gitlab url has changed. All of the urls have to be replaced with the new `git.pace.cloud`


docker-compose.yml
```diff
@@ -68,7 +68,6 @@ services:
       - JAEGER_SAMPLER_TYPE=const
       - JAEGER_SAMPLER_PARAM=1
       - JAEGER_SERVICE_NAME=testserver
-      - SENTRY_DSN=https://35223d89148145ae9d47ba3f268e6ece:1756d00edb714cf1aa07b478627176f3@sentry.jamit.de/193
```
alessandroMiceli, line 71: shouldn't we have pace.cloud?

pnull: Yes, we should. The authentication part will be wrong as well, most likely.. 🤔

flofuenf: This is a file for local testing. Local tests should not push sentries to our live environments, considering that other people outside of PACE can run tests randomly.

Ferlonas (approved):

PR #332, opened 2022-11-15
Add functions to check for Postgres & Redis connection failure errors

flofuenf:


arnold-iakab (approved):
:+1:

daemonfire300 (approved):

backend/postgres/errors.go
```diff
@@ -3,9 +3,39 @@
...
+		return true
+	}
+
+	// go-pg has similar check for integrity violation issues, here we check network issues
+	pgErr, ok := err.(pg.Error)
+	if ok {
+		code := pgErr.Field('C')
+		// We check on error codes of Class 08 — Connection Exception.
+		// https://www.postgresql.org/docs/10/errcodes-appendix.html
+		if code[0:2] == "08" {
```
daemonfire300, line 36: This might panic if len is insufficient, but I guess this would be a valid panci

mihailozarinschi: True, but it already checks on the Postgres error type, so there is already a guarantee that you only get error codes of this format.

PR #333, opened 2022-11-15
feat: add AddScope func to add a scope to token in context & tests

floaust:


daemonfire300:
@floaust did your tests pass locally, asking because CI is broken for forked PRs/

floaust:
@daemonfire300 yes they were successful

PR #338, opened 2023-01-04
Handle map with string slice values in json api implementation

arnold-iakab:
Defining AllowedPaymentClients and ForbiddenPaymentClients as map[string][]string is not possible according to the jsonapi implementation: '*poidbserviceclient.POI': "reflect.Set: value of type map[string]interface {} is not assignable to type map[string][]string".

### DIFF
--- a/http/jsonapi/request_test.go
+++ b/http/jsonapi/request_test.go
@@ -15,6 +15,8 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func TestUnmarshall_attrStringSlice(t *testing.T) {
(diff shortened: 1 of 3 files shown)


http/jsonapi/request_test.go
```diff
@@ -29,6 +29,7 @@ func TestUnmarshall_attrStringSlice(t *testing.T) {
 				"dec1": json.RawMessage(`"9.9999999999999999999"`),
 				"dec2": json.RawMessage("9.9999999999999999999"),
 				"dec3": json.RawMessage("10"),
+				"mss":  json.RawMessage(`{"field": ["string1","string2"]}`),
```
daemonfire300, line 32: this checks the happy path, could we also get a test that is expected to trigger the failure case or does Go's json unmarshal do magic converstion of JS Number to string by just wrapping "" around it, i.e.,

```json
`{"field": ["string1",1234, 9.9]}`
```
as test case.

daemonfire300 (changes requested):

http/jsonapi/request_test.go
arnold-iakab: added

PR #341, opened 2023-01-13
Refactor routine package so that it uses the redis lock implementation

arnold-iakab:


pkg/routine/lock.go
```diff
@@ -0,0 +1,220 @@
...
+		lockCtxChan <- ctx
+
+		keepUpLock(ctx, lock, l.lockTTL)
+		err := lock.Release()
+		if err != nil && err != redislock.ErrLockNotHeld {
+			log.Ctx(ctx).Debug().Err(err).Msgf("could not release lock %q", l.Name)
+		}
+	})
+
+	return <-lockCtxChan, cancel, nil
```
mihailozarinschi, line 137: The only thing preventing the "lock" logic to sit in its own package is the dependency on `routine.Run` to start this goroutine.

The biggest reason for using `routine.Run` is when the caller does not care about canceling the goroutine, but we still want to know within that goroutine when it is canceled (because the entire service is being shut down: it received a `syscall.SIGTERM`)

Since `AcquireAndKeepUp` controls the started goroutine (and also gives control to the caller to cancel it), we can simply replace it with the following:

```go
// Keep up lock, cancel lockCtx otherwise.
lockCtx, cancelLock := context.WithCancel(ctx)
go func() {
	defer pberrors.HandleWithCtx(ctx, fmt.Sprintf("keep up lock %q", l.Name)) // handle panics
	defer cancelLock()

	keepUpLock(lockCtx, lock, l.lockTTL)
	err := lock.Release()
	if err != nil && err != redislock.ErrLockNotHeld {
		log.Ctx(ctx).Debug().Err(err).Msgf("could not release lock %q", l.Name)
	}
}()

return lockCtx, cancelLock, nil
```

This gets rid of the circular dependency and allows us to put the lock logic in its supposed package.

daemonfire300: @mihailozarinschi @arnold-iakab can this be resolved?

arnold-iakab: Yes, the suggested changes were used and tested.

PR #343, opened 2023-02-15
Error generator

daemonfire300:


daemonfire300:
Original author @pieceofsoul

maintenance/errors/bricks.go
```diff
@@ -0,0 +1,91 @@
...
+func NewBricksError(opts ...BricksErrorOption) *BricksError {
+	e := &BricksError{}
+	for _, opt := range opts {
+		opt(e)
+	}
+	return e
+}
+
+func (e *BricksError) Error() string {
+	return e.code
```
mihailozarinschi, line 42: Maybe have a method `Code() string` that returns the `code`. And in `Error() string` return something including the detail, and maybe the title too, otherwise, neither the `title` nor the `detail` ar readable.

maintenance/errors/bricks.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (e *BricksError) Status() int {
+	return e.status
+}
+
+// AsRuntimeError - returns the BricksError as bricks runtime.Error which aligns
+// with a JSON error object
+func (e *BricksError) AsRuntimeError() *runtime.Error {
```
mihailozarinschi, line 51:
```suggestion
func (e *BricksError) ToRuntimeError() *runtime.Error {
```

maintenance/errors/bricks.go
```diff
@@ -0,0 +1,91 @@
...
+		Status: strconv.Itoa(e.status),
+		Code:   e.code,
+		Title:  e.title,
+		Detail: e.detail,
+	}
+	return j
+}
+
+// Equals - verifies if the given runtime.Error code equals the BricksError code
+func (e *BricksError) Equals(re *runtime.Error) bool {
```
mihailozarinschi, line 63: This method feels kind of useless, as usually comparisons are done via switch cases.

internal/service/generate/errordefinition/generator/generate.go
```diff
@@ -0,0 +1,116 @@
...
+		if jsonError.Code == "EXAMPLE" {
+			continue
+		}
+
+		httpStatusCode, err := strconv.Atoi(jsonError.Status)
+		if err != nil {
+			return "", fmt.Errorf("%w: unable to parse given http status code")
+		}
+
+		g.goSource.Func().Id(fmt.Sprintf("%s", jsonError.Code)).Params(
```
mihailozarinschi, line 96: Maybe the function name should be prefixed with `ERR_`, or `ErrAndCodeAsCamelCase()`??

It makes it clearer what that function is if it's prefixed like that.

mihailozarinschi (changes requested):

maintenance/errors/bricks.go
pieceofsoul: Getter functions have been added.

maintenance/errors/bricks.go
pieceofsoul: I added const definitions to be able to verify the generated application code strings, since this additionally removes the need to create an error object before being able to do a equal check against a runtime error code. This removes the need to have this equal function as well.

internal/service/generate/errordefinition/generator/generate.go
pieceofsoul: Add the proposed `ERR_` prefix.

Ferlonas (approved):

PR #351, opened 2023-06-29
fix pg pagination

mihailozarinschi:
Old behavior used to include the last item from the previously returned page because of `-1` on the offset. The existing tests did not cover this issue, new test cases were added to cover this.


pieceofsoul (approved):
Looks good to me.

PR #384, opened 2024-10-10
Extend linting

monstermunchkin:


srieckhof:
I'll wait with reviewing this PR until bun and sentry are completely finished.

maintenance/errors/context.go
```diff
@@ -21,7 +21,7 @@ func Hide(ctx context.Context, err, exposedErr error) error {
 
 	ret := err
 	if exposedErr != nil {
-		ret = fmt.Errorf("%w: %s", exposedErr, err)
+		ret = fmt.Errorf("%w: %s", exposedErr, err.Error())
```
srieckhof, line 24:
```suggestion
		ret = fmt.Errorf("%w: %w", exposedErr, err)
```

maintenance/errors/context_test.go
srieckhof, line 1: There is a mixture of verbs (%s, %w) I don't really understand - already in the original. Do you know why? How about just using %w?

PR #385, opened 2024-10-16
add sentry when a pod is becoming the active one

srieckhof:
This ticket refers to the problems we had with FSC, specifically when a pod crashed or two pods swapped active and passive roles without anybody noticing.

Relates to: https://git.pace.cloud/pace/web/service/fsc/-/issues/343

**Motivation:**
- let people know earlier that a new pod became active because this roles swap lead to a massive amount of offline stations that partially did not come online again (e.g. tankpool sites) 

**Changes description:**
 - refactor comments
 - elevate log levels from debug to info for some of the messages to better trace which pod gets which role assigned
 - add sentry after the log message telling about a new pod that became active

**NOTE:**
We will get the sentry also after a new deployment which is, however, no concern. Attention must only be paid once this sentry is risen without a deployment. Most likely this sentry will be the reason for the descision to restart the entire FSC service.
These changes will not fix the issue of many/some sites not getting online anymore. This will be part of a separate pull request.

### DIFF
--- a/maintenance/failover/failover.go
+++ b/maintenance/failover/failover.go
@@ -87,10 +87,8 @@ func NewActivePassive(clusterName string, timeToFailover time.Duration, client *
 	return ap, nil
 }
 
-// Run registers the readiness probe and calls the OnActive
-// and OnPassive callbacks in case the election toke place.
-// Will handle panic safely and therefore can be directly called
-// with go.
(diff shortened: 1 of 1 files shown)


monstermunchkin:
This is not needed, and it would add Sentry as a requirement for the `failover` package which is not ideal.

The `OnActive` function can be used to achieve the desired behavior.

PR #387, opened 2024-11-27
tracing: Fix bugs and clean up code

monstermunchkin:
This PR includes some code cleanup as well as two bug fixes, one in the routine functions, the other in the context transfer which lead to missing logs and scopes.

### DIFF
--- a/pkg/context/transfer.go
+++ b/pkg/context/transfer.go
@@ -19,7 +19,8 @@ import (
 // new context with all these objects.
 func Transfer(in context.Context) context.Context {
 	// transfer logger, log.Sink, authentication and error info
-	out := log.Ctx(in).WithContext(context.Background())
+	out := TransferTracingContext(in, context.Background())
+	out = log.Ctx(in).WithContext(out)
 	out = log.SinkContextTransfer(in, out)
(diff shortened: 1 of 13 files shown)


pkg/context/transfer.go
```diff
@@ -17,6 +17,7 @@ import (
 // Transfer takes the logger, log.Sink, authentication, request and
 // error info from the given context and returns a complete
 // new context with all these objects.
+// Deprecated: Use context.WithoutCancel instead.
```
arnold-iakab, line 20: I don't understand why this should be deprecated and instead we should user `context.WithoutCancel` 🤔

arnold-iakab (approved):

pkg/context/transfer.go
monstermunchkin: I believe this function was introduced because one wanted to remove the cancellations and timeouts from context. That's at least my understanding by reading the comments above some of those calls. However, since Go 1.21 there is `context.WithoutCancel` which does exactly that.

arnold-iakab: Hmm, ok makes.

PR #393, opened 2024-12-20
Refactor failover mechanism

srieckhof:
This is largely a refactoring in terms of improving comments, variable naming and logging. 

Regarding logging: some of the logs were changed from Debug() to Info() level in order to be able to see why certain requests, e.g. to redis, failed. 

The `lock.TTL(ctx)` is now hidden behind a function (`safeGetTTL()`) that recovers from a panic because I observed these kind of behaviors when testing another version of the failover improvement on dev.


monstermunchkin:
This is no longer needed.

PR #396, opened 2025-01-16
Fix failover

srieckhof:
This PR fixes the instability of FSC in failover mode. By removing the call to lock.TTL() we don't run into panics inside redislock anymore and, thus, don't quit the Run() method anymore. Without lock.TTL() one pod is active and the other two are passive. This even holds after the active pod is killed. Test fuelings still work.

Also, I removed the noisy "Redis query" debug message from being logged. Instead, only errors on command execution are logged.

### DIFF
--- a/maintenance/failover/failover.go
+++ b/maintenance/failover/failover.go
@@ -109,8 +109,7 @@ func (a *ActivePassive) Run(ctx context.Context) error {
 	tryRefreshLock := time.NewTicker(a.timeToFailover)
 
 	// Ticker to try to acquire the lock if in passive or undefined state
-	tryAcquireInterval := 500 * time.Millisecond
-	tryAcquireLock := time.NewTicker(tryAcquireInterval)
+	tryAcquireLock := time.NewTicker(500 * time.Millisecond)
 
 	for {
 		select {
--- a/maintenance/failover/failover.go
+++ b/maintenance/failover/failover.go
@@ -155,23 +155,8 @@ func (a *ActivePassive) Run(ctx context.Context) error {
 				logger.Debug().Msg("becoming active")
 				a.becomeActive(ctx)
 
-				// we are active, renew if required
-				d, err := lock.TTL(ctx)
-				if err != nil {
-					logger.Debug().Err(err).Msg("failed to get TTL")
-				}
-				if d == 0 {
-					// TTL seems to be expired, retry to get lock or become
-					// passive in next iteration
-					logger.Debug().Msg("ttl expired")
-					a.becomeUndefined(ctx)
-				}
-				refreshTime := d / 2
-
-				logger.Debug().Msgf("set refresh to %v", refreshTime)
-
 				// set to trigger refresh after TTL / 2
-				t.Reset(refreshTime)
+				t.Reset(a.timeToFailover / 2)
 			}
 		}
 	}
--- a/maintenance/failover/failover.go
+++ b/maintenance/failover/failover.go
@@ -18,8 +18,6 @@ import (
 	"github.com/redis/go-redis/v9"
 )
 
-const waitRetry = time.Millisecond * 500
-
 type status int
 
 const (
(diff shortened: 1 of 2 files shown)


maintenance/failover/failover.go
```diff
@@ -110,68 +105,50 @@ func (a *ActivePassive) Run(ctx context.Context) error {
...
-					a.becomeUndefined(ctx)
-				}
-				refreshTime := d / 2
-
-				logger.Debug().Msgf("set refresh to %v", refreshTime)
-
-				// set to trigger refresh after TTL / 2
-				t.Reset(refreshTime)
+				// Reset the refresh ticker to half of the time to failover
+				tryRefreshLock.Reset(a.timeToFailover / 2)
```
srieckhof, line 151: Just FYI: We are setting the refreshLock timer to 5s (if timeToFailover=10s) to give enough time for the active pod to refresh the lock in time - even considering network/keydb delays or outages lasting less than 5s. Everything above 5s may lead to the active pod not being able to refresh the lock in time.

maintenance/failover/failover.go
```diff
@@ -110,68 +105,50 @@ func (a *ActivePassive) Run(ctx context.Context) error {
...
-	// t is a ticker that reminds to call refresh if
-	// the token was acquired after half of the remaining ttl time
-	t := time.NewTicker(a.timeToFailover)
+	// Ticker to try to refresh the lock's TTL before it expires
+	tryRefreshLock := time.NewTicker(a.timeToFailover)
 
-	// retry time triggers to check if the look needs to be acquired
-	retry := time.NewTicker(waitRetry)
+	// Ticker to try to acquire the lock if in passive or undefined state
+	tryAcquireInterval := 500 * time.Millisecond
```
arnold-iakab, line 112: I would either make it a constant or directly use `500 * time.Millisecond` in the NewTicker function bellow.

arnold-iakab (commented):
LGTM, just some small suggestions.

maintenance/failover/failover.go
arnold-iakab: Sorry, I cannot follow, how do we give more time to the active pod (which is the one that is reaching this part of the code) by setting the ticker to tick sooner?

srieckhof: No problem. Once the lock has been successfully obtained with a TTL=timeToFailover (TTL = 10 s) the pod becomes active (should only take some milliseconds executing OnActive()) and then resets the refreshLock timer from formerly 10 s to 5 s. Then, every 5 s the active pod will be able to refresh the lock by resetting its TTL to 10 s (see lock.Refresh()). Actually, the active pod could also try to refresh the lock after 9 s or so which is however just before the lock would expire. So, in order to account for network outages/delays (network between pod and keydb) or outages of keydb itself (ok, in this case the lock/key may be gone entirely anyway, that is, in case of a keydb crash/restart) we try to refresh the lock well within the 10 s TTL. To do so we use half of the TTL or half of the time to failover (TTL/2 is approx. timeToFailover/2) for the refresh ticker.

maintenance/failover/failover.go
srieckhof: I remove this constant `waitRetry` (see above). For me it was no use setting a constant when used only once. I wanted to set it close to the ticker definition and wanted to give it an own variable to be very descriptive about what it is.

maintenance/failover/failover.go
arnold-iakab: Got it. For some reason yesterday I was seeing `tryAcquireLock` being reset instead of `tryRefreshLock`.

As a side note, the amount of time it takes for `OnActive()` to execute is controlled by the code that uses the failover mechanism. In case of `fsc` indeed it should not take much since it iterates over a list of functions that are run in routines.

arnold-iakab (approved):

maintenance/failover/failover.go
monstermunchkin: I agree with @arnold-iakab.

> For me it was no use setting a constant when used only once.

Well, the same goes for the new variable you introduced.

I'd prefer keeping the constant, and adding a comment.

monstermunchkin (changes requested):

maintenance/failover/failover.go
srieckhof: That's fine. I put the duration into the ticker creation.

monstermunchkin (approved):

PR #401, opened 2025-02-07
middleware: Use map for external dependencies

monstermunchkin:
This replaces the dependency slice with a map. The duration of external dependencies is now accumulated instead of appended to the slice.


http/middleware/external_dependency_test.go
```diff
@@ -56,6 +56,10 @@ func Test_ExternalDependencyContext_String(t *testing.T) {
 
 	edc.AddDependency("test4", time.Millisecond*123)
 	assert.EqualValues(t, "test1:1,test2:0,test3:1000,test4:123", edc.String())
+
+	// This should overwrite the previous value
```
srieckhof, line 60: I think this comment doesn't fit to what the code actually does. The value is added but doesn't overwrite.

srieckhof:
Other than the above comment everything looks good.

srieckhof (approved):

PR #402, opened 2025-02-18
grpc: Return external dependencies to client

monstermunchkin:
External dependencies which are updated during the request life-cycle need to be sent back to the client.


grpc/middleware.go
```diff
@@ -36,3 +37,22 @@ func EncodeContextWithUTMData(parentCtx context.Context) context.Context {
 	}
 	return metadata.AppendToOutgoingContext(parentCtx, utmMetadataKey, w.String())
 }
+
+// AddExternalDependencyMetadataToContext adds external dependencies to the context if they are present in the metadata.
+func AddExternalDependencyMetadataToContext(ctx context.Context, md metadata.MD) context.Context {
```
arnold-iakab, line 42: I am a bit confused, why is the function added? Who and where is this supposed to be used?

arnold-iakab: Not confused anymore.

arnold-iakab (approved):
lgtm

grpc/middleware.go
```diff
@@ -7,6 +7,7 @@ import (
 	"encoding/gob"
 	"strings"
 
+	"github.com/pace/bricks/http/middleware"
```
srieckhof, line 10: I think that, architecturally speaking, it's not good to import http/middleware in grpc/middleware. It obviously works now but there is the danger of cyclic dependencies in the future. I suggest to do a (hopefully) small refactoring of extracting the e.g. `middleware.ExternalDependencyContextFromContext` into an own separate package that is used by both middleware packages.

monstermunchkin: > I think that, architecturally speaking, it's not good to import http/middleware in grpc/middleware. It obviously works now but there is the danger of cyclic dependencies in the future.

Should that ever be the case, we can gladly refactor it. For now though, I won't change it because it will break backward compatibility as it's an exported functions which is imported in other packages.

srieckhof: Ok, I see. Then, it's fine for me. Thanks.

srieckhof (approved):

PR #403, opened 2025-02-28
make enabling sentry tracing configurable; default is true

srieckhof:
This adds a new environment variable that makes it possible to disable sentry tracing for a specific service. Tracing is enabled per default.


monstermunchkin (approved):

monstermunchkin (changes requested):
Could you please mention the `SENTRY_ENABLE_TRACING` variable in the README?

monstermunchkin (approved):

